PR #7, opened 2018-10-22
Minimal travis support

### LINKED ISSUE #3: Add CI for the project
There are a few CI you can choose from like Travis.

Kokan:
This is just something very basic.
* there was a nasty dependency on `$GOPATH` environment variable, just plain simple removed it
* creates and maps current dir `bin/` where the executables going to be saved (the mkdir part maybe not ideal)
* added `.gitignore` for the created files
* build status logo for master

Fixes #3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+bin/
+pkg/crd/apis/danm/*/zz_generated.deepcopy.go
+pkg/crd/client/
+pkg/vendor/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+sudo: required
+
+language: ruby
+
+services:
+  - docker
+
+env:
+  - GOPATH=/home/travis/gopath
+
+script:
+  - mkdir -p $GOPATH/{bin,pkg}
+  - mkdir -p $GOPATH/src/github.com/nokia/danm/
+  - mv * $GOPATH/src/github.com/nokia/danm/
+  - cd $GOPATH/src/github.com/nokia/danm/
+  - ./build_danm.sh
+
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # DANM
+[![Build Status](https://travis-ci.org/Nokia/danm.svg?branch=master)](https://travis-ci.org/Nokia/danm)
 <img src="https://github.com/nokia/danm/raw/master/logo.png" width="100">
 
 ## Table of Contents


Kokan:
In my fork I tried it out: https://travis-ci.org/Kokan/danm/builds/444607190

Levovar:
Generally cool, but that "nasty piece of dependency" is actually Golang standard way to mark your workspace :)
Removing it would mean the whole building works only if the user starts the container from the project's directory
Same for the executables: they are supposed to be put into the host's GOPATH/bin directory by default

Dependency on the presence of this env vár should remain until we migrate to the vgo / go modules based dependency management, simply because this is how go works by default

Kokan:
Sure I know a little about go :)

`Dependency on the presence of this env vár should remain until we migrate to the vgo / go modules based dependency management, simply because this is how go works by default`
I think this reasoning is not correct; you completely ignore the build container, I mean go works with those env variables, and it works fine inside the build container/environment, where the build related information should stay and not leak outside. This way your build environment is not really encapsulated, because everything you have locally in the `$GOPAHT` would poison your builds :)

But I can drop it as there are many thing that must be discussed with the build and deployment, so I hope eventually we are going to get there.

Levovar:
"everything you have locally in the $GOPAHT would poison your builds :)"
that is the whole idea of how traditionally depencies were managed in Golang :)

You have a workspace, you go get Github Golang projects which will be put into your $GOPATH/src/github.com..., you compile them, dependencies will also come from your $GOPATH/src/..., and whatever you build is put into $GOPATH/bin

What we simulate within the container is exactly this, as if it would happen on the host when the developer issues "go install ....danm", or "go build ....danm".

Now, I'm okay if you remove $GOPATH usage from the build script as long as the following requirements are satisfied:
- the project itself cannot have directories, files, anything added to it which are purely needed to build it (for example an empty /bin directory etc.)
- the compiled binaries are put into the developer's workspace's bin directory, as it should be if she would have invoked a simple "go build" command

Levovar:
Cool. And thank you for the change!

yeah the generators are kind of special, but in short we need a very specific version from them, otherwise the whole project does not work :)
Generally speaking the whole language does not handle this case when parts of the code are generated by binaries coming also from other Golang project... this is the main reason of providing build automation in the scope of the project, rather than just pushing the responsibility of doing all these steps to the user

Levovar:
Is the link put into the README good though?

Or maybe I don't understand how this integration should work :) 
a little overview would be appreciated: when a travis is triggered from the parent repo for example. Or we need to separately set that up?

Kokan:
@Kokan sorry, it is not. (capital N in nokia is wrong).

Kokan:
@Levovar this should fix it: https://github.com/nokia/danm/pull/8

Kokan:
The way it works (the new version of Travis at least), it adds a webhook to your repository (you can check it  via Settings/hooks), that hook is triggered when change happens in your repository (PR, commit, ...).

Travis starts the build on each branch in the original repository, and in the fork if a PR is created from them.

If you have a fork, and create a branch you do not have a Travis by default for it, you have to enable in travis for your fork. But in case of a PR, it get executed.

When Travis ends execution it sends the result back to github, which can be reported back to the commit.

(you could find better description I guess in travis own site)

Levovar:
yes, thanks for the quick recap anyway!

PR #8, opened 2018-11-01
Fix travis link in readme.md

Kokan:
Signed-off-by: Kokan <kokaipeter@gmail.com>


Levovar:
At least we test the travis integration too :)

Kokan:
Yes, but it was also executed for the previous PR. Btw there are also tricks like `[skip-ci]` in title, that could ask Travis not to execute.

PR #10, opened 2018-11-07
To correct some small mistakes

clivez:
Correction for some small mistakes


Levovar:
the funny thing is that I really use this exact DanmNet CRD spec in my real cluster, and K8s never said a damn word about the typo :)

Thanks!

PR #13, opened 2018-11-10
Integrate MACVLAN CNI to DANM with dynamic integration level + Extend IP route provisioning support to ALL backends

Levovar:
This will enable the creation of MACVLAN interfaces based on the dynamic DanmNet API.
The CNI used to create the interfaces is the basic CNI coming from:
https://github.com/containernetworking/plugins/blob/master/plugins/main/macvlan/macvlan.go

Interfaces are created in bridge mode, with 1500 MTU.

Change is not yet tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,8 +279,8 @@ This way users can dynamically configure various networking solutions via the sa
 A generic framework supporting this method is built into DANM's code, but still this level of integration requires case-by-case implementation.
 As a result, DANM currently supports two integration levels:
 
- - **Dynamic integration level:** CNI-specific network attributes (such as IP ranges, master host devices etc.) can be controlled on a per network level, taken directly from a DanmNet object
- - **Static integration level:** CNI-specific network attributes (such as IP ranges, master host devices etc.) can be only configured on a per node level, via a static CNI configuration files (Note: this is the default CNI configuration method)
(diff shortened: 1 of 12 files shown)


Levovar:
setupHost in net.go should be adjusted to also handle host interface provisioning for macvlan backend

Levovar:
The review contains the generalization of how IP route provisioning is handled within DANM.

Previously IP routes were only provisioned into a container's netNS if the NEtworkType of the related DanmNet was "ipvlan".

The related code was refactored to be generic, and is now invoked from cnidel after a delegated CNI created an interface.
As a result, DANM can now provision both generic, and policy-based IP routes for both static, and dynamic-level backends.

Actually there is not much separating the two after this change :)

Levovar:
Change is now functional tested, all newly added functionalities work.
After this review it will be possible to:
- provision MACVLAN interfaces purely from the DanmNet API
- all supported options work with MACVLAN too, IPv6, static IP allocation, routes, proutes, vlan, and vxlan included
- all backends, both dynamic and static can now have default IP routes provisioned based on the DanmNet API
- all backends, both dynamic and static can now have Pod-level policy-based IP routes provisioned into their routing tables

Only documentation remains, and we are good to go!

Issue #11, opened 2018-11-07
example directory missing


clivez:
A demo about Multi-domain service discovery is mentioned in the README, but the example directory does not exist in current project, will someone please upload the related files into the project? Thank you.

Levovar:
yep, it is definitely on me to provide these! 

Unfortunately I didn't have much time recently to finish the README, but nevertheless it is in my backlog. I will try and work out the example manifest for the scenario described in the README within week, let's say

In case I'm late feel free to nag me more about it :)

PR #16, opened 2018-11-12
Adding svcwatcher demo scenario

Levovar:
Fixes https://github.com/nokia/danm/issues/11 (@clivez )
Manifest files can be found under example/svcdemo

Manifest files should be instantiated in this order -> Namespaces -> DanmNets -> Deployments -> Services

Later on I might also add a short video showcasing the demo. README is not yet updated.


clivez:
Thank you @Levovar.
I'm trying the DANM these days, maybe there will also be some other things that may bother you, thank you in advance.

Levovar:
> Thank you @Levovar.
> I'm trying the DANM these days, maybe there will also be some other things that may bother you, thank you in advance.

sure, any time :)

PR #17, opened 2018-11-14
Rename the former watcher as netwatcher

clivez:
I guess you may want to use the name "netwatcher" instead of "watcher", but currently it is still "watcher" in the code, so I made a modification to make it compatible with what mentioned in README and integration/


Levovar:
Yes indeed, thank you!

First there was only "watcher", hence the generic name. Once we have added svcwatcher we started calling the original watcher "netwatcher", just to separate the two. But yeah, code level renaming was not done yet :)

PR #26, opened 2018-11-26
issue24: precheck

rospring:


Levovar:
Looks good to me, just delete the commented out lines :)

Levovar:
Stuff is tested, stuff looks to be working!

Actually I think this change also noticeably increases instantiation performance

Issue #15, opened 2018-11-12
Refactor "container_prefix" DanmNet attribute


Levovar:
It should not be mandatory.

Instead, we should revert to the original solution, where an invocation ID parameter was passed to network setup functions.
If container_prefix is not provided, interfaces are set-up with "eth"+INVOCATION_ID

Levovar:
Enhancement will also include supporting "container_prefix" attribute even for statically configured / unmanaged CNIs.

PR #27, opened 2018-11-30
Introduce dynamic container interface naming as a fallback option

Levovar:
Implements solution to https://github.com/nokia/danm/issues/15
Container_prefix API attribute became optional.
A NIC connecting to a network having it set will continue to be name accordingly.
A NIC connecting to a network not having it set will be dynamically provisioned
a Pod-unique name in Pod Instantiation time.
Enhancement works for in-built IPVLAN, static delegates, and dynamic delegates!

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ danm        **fakeipam**      host-device  ipvlan       macvlan      ptp        
  ```
 kubectl create -f integration/manifests/netwatcher/
 ```
-Note: we assume RBAC is configured for the Kubernetes API, so the manifests provides the required Role and ServiceAccount for this case.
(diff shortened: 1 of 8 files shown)


Levovar:
TODO1 (reminder for myself :) ) : when Options is provided for a static network only because of container_prefix, host_device and rt_tables are also currently mandated on the schema-level
needs to be removed

Levovar:
TODO2: actually the proposed solution does not work :) CNI_IFNAME is an environment level parameter, and is not part of the streamed-in netconf
solution needs to be change to directly invoke invoke.ExecPluginXY with manually assembled environment variable-set per delegation

PR #29, opened 2018-12-03
RRAC for net and svcwatcher

libesz:
This PR avoids the usage of hostpath mounts on every node and uses RBAC for the two Danm daemonsets. It allows in-cluster config for kube clients smoothly. The needed role content is reverse engineered and seems working, but please take a look, might be too much or too less still 😇.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ You are now ready to use the services of DANM, and can start bringing-up Pods wi
 
  **+1. OPTIONAL: Create the servicewatcher DaemonSet by executing the following command from the project's root directory:**
  ```
-kubectl create -f integration/manifests/svcwatcher/svcwatcher_ds.yaml
+kubectl create -f integration/manifests/svcwatcher/
 ```
  This component is an optional part of the suite. You only need to install it if you would like to use Kubernetes Services for all the network interfaces of your Pod.
 Note: svcwatcher already leverages DANM CNI to create its network interface. Don't forget to change the name of the network referenced in the example manifest file to one which:
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ danm        **fakeipam**      host-device  ipvlan       macvlan      ptp        
  ```
 kubectl create -f integration/manifests/netwatcher/
 ```
-Note: don't forget to change the names of files and directories pointing to valid kubeconfig files, and TLS certificates used by the K8s API server in your infrastructure before instantiating the component! 
+Note: we assume RBAC is configured for the Kubernetes API, so the manifests provides the required Role and ServiceAccount for this case.
 
 You are now ready to use the services of DANM, and can start bringing-up Pods within your cluster!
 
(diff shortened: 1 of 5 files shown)


integration/manifests/svcwatcher/0svcwatcher_rbac.yaml
```diff
@@ -0,0 +1,66 @@
...
+  - list
+  - watch
+  - get
+  - update
+  - create
+  - delete
+- apiGroups:
+  - ""
+  resources:
+  - services
```
Levovar: svcwatcher also needs DanmNet,DanmEp, and EndPoints as far as I know

README.md
```diff
@@ -171,15 +171,15 @@ danm        **fakeipam**      host-device  ipvlan       macvlan      ptp
 
  **7. Create the netwatcher DaemonSet by executing the following command from the project's root directory:**
  ```
-kubectl create -f integration/manifests/netwatcher/netwatcher_ds.yaml
+kubectl create -f integration/manifests/netwatcher/
 ```
 Note: don't forget to change the names of files and directories pointing to valid kubeconfig files, and TLS certificates used by the K8s API server in your infrastructure before instantiating the component!
```
Levovar: we should probably update other references throughout the document implying kubeconfig based setup

or better, we should describe somewhere that on the code-level both kubeconfig based and RBAC / SA based incluster configuration is supported, but the example describes the latter

integration/manifests/svcwatcher/0svcwatcher_rbac.yaml
libesz: I think we have those also in this ClusterRole. Danm stuff starting in line 37 with full control and also the endpoints  above in line 19

Levovar: ah my bad, I still didn't have my morning coffee :)

README.md
Levovar: on a second thought: I'm also fine supporting only RBAC fro netwatcher too. Anyway this should be done for production environment in most cases
but then we should delete the kubeconf option from the netwatcher code

libesz: Either is fine by me, you choose 😄. However if we remove the kubeconfig option, we break BW compatibility. It might not hurt anybody, but still.

Levovar: yeah it's BW incompatible, but also the old solution is not good anyway

I think RBAC is "default" enough that we can do it

integration/manifests/netwatcher/0netwatcher_rbac.yaml
```diff
@@ -0,0 +1,42 @@
...
+kind: ClusterRole
+metadata:
+  labels:
+    kubernetes.io/bootstrapping: rbac-defaults
+  name: system:netwatcher
+rules:
+- apiGroups:
+  - "danm.k8s.io"
+  resources:
+  - danmnets
```
Levovar: I think Pod will be also needed. CNI reads the Pod for its annotation

libesz: Buuuuut.... this one is only for netwatcher.
CNI is configured through the /etc/cni/net.d config for kubelet, or not? For containerized kubelet, it also uses the local /etc/kubernetes/kubelet.conf and not ServiceAccounts, I would think.

Levovar: yes, I'm just being stupid again

PR #34, opened 2018-12-12
Default net

### LINKED ISSUE #21: Proposal: default network
The use-case I would like to raise here is to have a chance to successfully deploy applications without danm related annotations. This pretty much involves all common reusable components out there (i.e. community maintained helm charts, kubernetes addons like coredns or dashboard, etc.) and usually solved with other (more simple) CNI plugins.
If we could enhance danm to have some configuration option as a "default" network connectivity in certain namespaces, than it could be used for the Pods in the same ns which does not require any specific network access. Default danmnets could have some restrictions as needed.

libesz:
**What type of PR is this?**
feature

Admin is now able to provision a (single) default network (a danmnet resource with name: "default") in the target kubernetes namespace, which will be used when there is nothing defined in the annotation section of a Pod/Controller.

Fixes #21 

```release-note
Admin is now able to provision a (single) default network (a danmnet resource with name: "default") in the target kubernetes namespace, which will be used when there is nothing defined in the annotation section of a Pod/Controller.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,6 +264,7 @@ When network management is delegated to CNI plugins with static integration leve
 Pods can request network connections to DanmNets by defining one or more network connections in the annotation of their (template) spec field, according to the schema described in the **schema/network_attach.yaml** file.
 
 For each connections defined in such a manner DANM will provision exactly one interface into the Pod's network namespace, according to the way described in previous chapters (configuration taken from teh referenced DanmNet API object).
+Note: if the Pod annotation is empty (no danmnet connection is defined), DANM will fall back to use a default network definition, if created. The danmnet object in the target K8S namespace must use the name: "default". It can have any network type (either delegated or ipvlan). Naturally, user in this case cannot specify any further property for the Pod (i.e. static IP address). This way, the user is able to use unmodified manifest files (i.e. community maintained Helm charts or Pods created by K8S operators).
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -268,6 +268,7 @@ For each connections defined in such a manner DANM will provision exactly one in
 In addition to simply invoking other CNI libraries to set-up network connections, Pod's can even influence the way their interfaces are created to a certain extent.
 For example Pods can ask DANM to provision L3 IP addresses to their IPVLAN or SRI-OV interfaces dnyamically, statically, or not at all!
 Or, creation of policy-based L3 IP routes into their network namespace is also a supported by the solution.
+Note: if the Pod annotation is empty (no danmnet connection is defined), DANM will fall back to use a default network definition, if created. The danmnet object in the target K8S namespace must use the name: "default". It can have any network type (either delegated or ipvlan). Naturally, user in this case cannot specify any further property for the Pod (i.e. static IP address). This way, the user is able to use unmodified manifest files (i.e. community maintained Helm charts or Pods created by K8S operators).
```
Levovar: could we move this to one paragraph higher, i.e. starting from line 267?

It would make sense to talk about the fallback option right after we detailed how can a Pod define network connections

libesz: Done.

PR #45, opened 2019-02-11
Comment and figure out how to move forward

### LINKED ISSUE #35: Danm installer and popularization
feature

**What happened**:
Installation is currently a manual procedure, which might include also the compilation of the components.

Since the installation steps are not depending on too much decisions or mandatory admin interactions, it could and should be automated similarly to other CNI plugins.

**Draft specs**:
Install and start up all danm components with a single (or maybe a few) `kubectl apply` command. Controller/watcher manifests and RBAC are already available, but the CNI plugin itself has to be distributed also with a DaemonSet+host mounts. I would assume the kubelet is either running natively or if in container, than it has the relevant CNI directories mounted from host.
The installation optionally should cover a default network provisioning for the kube-system namespace in order to allow CoreDNS and other basic addons to work.
The feature depends on #18. The closing task should be to contribute the installation steps to the K8S Kubeadm documentation [here](https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/) along with the integration to [KubeSpray](https://github.com/kubernetes-sigs/kubespray).

dromie:
**What type of PR is this?**
>
> bug
> cleanup
> design
> documentation
> failing-test
 feature

Fixes #35

**Does this PR introduce a user-facing change?**: NONE

```release-note

```


build_docker.sh
```diff
@@ -1,12 +1,45 @@
-#!/bin/sh
-
+#!/bin/bash -xe
+version=3.1.0
 echo 'Building DANM builder container'
-docker build --no-cache --tag=danm_builder:1.0 build/
+docker build --build-arg http_proxy=$http_proxy --build-arg https_proxy=$https_proxy --tag=danm_builder:1.0 build/
 
 echo 'Running DANM build'
-docker run --rm --net=host --name=danm_build -v $GOPATH/bin:/go/bin -v $GOPATH/src:/go/src danm_builder:1.0
```
Levovar: 1: removal of -rm means, that whenever there is a failure during the execution of this build script, the build container instance won't clean itself

2: any particular reason of removing the src and bin directory mount? you could simply copy from "/go/bin/XY" in lines 10-15, rather than needing to do docker cp from outside of the container

integration/deployment/default.yaml
```diff
@@ -0,0 +1,27 @@
...
+apiVersion: danm.k8s.io/v1
+kind: DanmNet
+metadata:
+  name: default
+  namespace: kube-system
+spec:
+  NetworkID: default
+  NetworkType: flannel
+---
+apiVersion: danm.k8s.io/v1
```
Levovar: we only need the default networks in the two namespaces, no need for the explicit Flannel network definition

integration/deployment/Dockerfile
```diff
@@ -0,0 +1,7 @@
+#target: danm-deployer
+FROM alpine:3.8
+MAINTAINER Nandor Galambosi <nandor.galambosi@nokia-bell-labs.com>
+COPY .cache/kubectl /opt/bin/
+COPY bin/* /opt/bin/
```
Levovar: svcwatcher and netwatcher binaries you don't need to install into /opt/bin on the target
I guess they are currently added to the installer container, unnecessarily

integration/deployment/deploy.sh
```diff
@@ -0,0 +1,39 @@
...
+cp /opt/bin/* /target/cni-bin/
+cat >/target/netd/00-danm.conf <<-EOF
+{
+  "name": "meta_cni",
+  "type": "danm",
+  "kubeconfig": "$DANMKUBECONFIG"
+}
+EOF
+NAME=kubernetes
+HOST=$KUBERNETES_SERVICE_HOST
```
Levovar: HOST and PORT variables I don't see defined in the initcontainer part of netwatcher manifest
don't we need it?

integration/deployment/prebuild.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+mkdir -p .cache
+if [ ! -f .cache/kubectl ];then
```
Levovar: this might be a tough ask, but... can't we simply use either a Python, or a Golang Kubernetes client in a programmable manner, rather than relying on the presence of the kubectl binary?
I think you only use it create the CRDs, and the 2 DanmNet manifests? that could be like 50 lines in either language, but minus one dependency

integration/deployment/default.yaml
```diff
@@ -0,0 +1,27 @@
...
+kind: DanmNet
+metadata:
+  name: default
+  namespace: kube-system
+spec:
+  NetworkID: default
+  NetworkType: flannel
```
Levovar: where is the Flannel DaemonSet, and the Flannel network config installed? I can't find it in the review

either we should take care of installing everything on our own, or we should not configure Flannel as the default backend

integration/manifests/netwatcher/0netwatcher_rbac.yaml
```diff
@@ -17,11 +17,22 @@ rules:
...
+  - danmeps
   verbs:
   - get
   - list
   - watch
   - update
+  - create
+- apiGroups:
+  - ""
+  resources:
```
Levovar: I don't think either the netwatcher, or the installer needs pods, or services

integration/deployment/Dockerfile
Levovar: internally we have discussed installing the CNI binaries of all backends, which we support with the dynamic level
These are (in addition to danm):
- macvlan
- sriov (specific version for the time being

Source from where you can get these:
- macvlan: https://github.com/containernetworking/plugins/releases/download/v0.7.0/cni-plugins-amd64-v0.7.0.tgz  (version could be a config parameter ofc)
- sriov: this one we build within our product, but here is the code for reference:
&&   mkdir -p ${GOPATH}/src/github.com/intel/sriov-cni \
&&  curl -fsSL -k https://github.com/intel/sriov-cni/archive/v0.2.tar.gz | tar zx --strip-components=1 -C ${GOPATH}/src/github.com/intel/sriov-cni \
&&  cd ${GOPATH}/src/github.com/intel/sriov-cni \
&&  ./build \
&&  mv ${GOPATH}/src/github.com/intel/sriov-cni/bin/sriov /opt/cni/bin/sriov \
right now we need the specific, 0.2, or latest 0.4 version

+1: if you decide to go the "fully integrated Flannel" way, you can get its CNI from here: 
go get -v github.com/containernetworking/plugins/plugins/meta/flannel

integration/deployment/deploy.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh -e
+DANMKUBECONFIG=/etc/kubernetes/danm.kubeconfig
+cp /opt/bin/* /target/cni-bin/
```
Levovar: the script needs to start with checking whether danm is already installed on the cluster, and should exit immediately if it is.
we should not overwrite existing installations

checking the presence of danm CNI, and 00-danm.conf is suitable

integration/deployment/default.yaml
dromie: The current svcwatcher_ds explicitly requires flannel network.

integration/deployment/Dockerfile
dromie: This container is used only as an initContainer. And initContainer will install all binaries to kubernetes nodes local filesystem.

integration/deployment/default.yaml
Levovar: the accurate statement would be "the current svcwatcher requires the presence of A network connection through which it can reach the K8s API server, and because Levo uses Flannel in its Nokia internal product for the purpose, it was accidentally left in its example manifest file" :)

integration/deployment/Dockerfile
Levovar: yep, and my comment is that svcwatcher and netwatcher binaries are not required to be installed on the local filesystem
these binaries are only needed within their own container image, which we build separately

only danm, and fakeipam are required to be copied to the filesystem for Kubelet

build_docker.sh
dromie: Ok, I'll rewrite it.

build_docker.sh
```diff
@@ -1,12 +1,45 @@
-#!/bin/sh
-
+#!/bin/bash -xe
+version=3.1.0
 echo 'Building DANM builder container'
-docker build --no-cache --tag=danm_builder:1.0 build/
```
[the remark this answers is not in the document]
dromie: bug

build_docker.sh
```diff
@@ -1,12 +1,45 @@
-#!/bin/sh
-
+#!/bin/bash -xe
+version=3.1.0
```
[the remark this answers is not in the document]
dromie: yep.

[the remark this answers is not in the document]
dromie: Although.....current manifests already contains hardcoded versions....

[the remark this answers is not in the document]
Levovar: how about removing hard-coded version from the example manifests, substitute them with a TEMPLATE string, and sed TEMPLATE string in build time based on the command line parameter?

integration/deployment/prebuild.sh
dromie: kubectl is not required anymore, will be removed.

integration/deployment/Dockerfile
dromie: Ok, only danm, and fakeipam will be copied

integration/manifests/netwatcher/0netwatcher_rbac.yaml
dromie: Without these permissions I get this error:
default   2019-02-14 13:51:18 +0100 CET   2019-02-14 13:51:18 +0100 CET   1         test-55678dddc4-xfwdj.15833c02ad922285   Pod                 Warning   FailedCreatePodSandBox   kubelet, node2.testdhn.local   Failed create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container "341f11a5b668ba357e933bd9c1980d5dae6e06045708db275e6fd83f25540d02" network for pod "test-55678dddc4-xfwdj": NetworkPlugin cni failed to set up pod "test-55678dddc4-xfwdj_default" network: Pod manifest could not be parsed with error: failed to get Pod info from K8s API server due to:pods "test-55678dddc4-xfwdj" is forbidden: User "system:serviceaccount:kube-system:netwatcher" cannot get resource "pods" in API group "" in the namespace "default"

Levovar: why are you using the netwatcher ServiceAccount for the DANM CNI, though?
it should use its own kubeconfig

dromie: It gets it own kubeconfig :) 
Would you like me to create a separate serviceAccount for danm with the required roles?
Installation guide does not really specify this properly.....

Levovar: Kubeconfig and ServiceAccount are kind of mutually exclusive.
Practically speaking the log suggests the DANM CNI was initalized from InCluster config in your environment, and not from its kubeconfig.
on second thought you kind of "forced" this InCluster initialization from the kubeconfig you provided to the CNI. but you can simply state a username in your kubeconfig, and embed the certificates to be used into the kubeconf of DANM

from security POV the CNI definitely needs to have its own role, whether it is configured through InCluster config, or through embedded certs

Levovar: (own role, together with its own user)

Levovar: For a reference, our kubeconfig in the product looks liek this:
[cloudadmin@controller-1 ~]$ sudo cat /etc/kubernetes/kubeconfig/danmc.yml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: <BASE64_ENCODED_CA_CERT>
    server: https://10.254.0.1:443
  name: kubernetes
contexts:
- context:
    cluster: kubernetes
    user: danm
  name: default
current-context: default
kind: Config
preferences: {}
users:
- name: danm
  user:
    client-certificate-data: <BASE64_ENCODED_CLIENT_CERT>
    client-key-data: <BASE64_ENCODED_CLIENT_KEY>

Levovar:
Are we still doing this? :)

Levovar:
somewhat invalidated by https://github.com/nokia/danm/pull/196
any additional PRs aiming to further automate DANM deployment will be also based on 196, making this one outdated

PR #50, opened 2019-02-25
Integrating sr iov dp

peterszilagyi:
**What type of PR is this?**

feature

DANM Integration with Intel SR-IOV Network Device Plugin and latest Intel SR-IOV CNI
**Which issue(s) this PR fixes** _(in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)_:
Fixes #

```
action required
*DanmNet definition's device_pool attribute should contain the SR IOV DP ResourceName 
* Pod definitions having sriov DanmNets should have SR-IOV DP resource requests too 
```


pkg/danm/danm.go
```diff
@@ -183,13 +191,115 @@ func extractConnections(args *cniArgs) error {
   return nil
 }
 
+func getResourcePrefix(args *cniArgs, resourceType string)(string,error){
+  confArgs, err := loadNetConf(args.stdIn)
```
Levovar: this is already done here :https://github.com/nokia/danm/blob/0ea0698fbd56252f8d303c58283b2e79c26f0a82/pkg/danm/danm.go#L103

pkg/danm/danm.go
```diff
@@ -183,13 +191,115 @@ func extractConnections(args *cniArgs) error {
...
+  if err != nil {
+    return "", errors.New("cannot load CNI NetConf due to error:" + err.Error())
+  }
+  k8sClient, err := createK8sClient(confArgs.Kubeconfig)
```
Levovar: could you re-use the same client we are creating for reading the Pod info: https://github.com/nokia/danm/blob/0ea0698fbd56252f8d303c58283b2e79c26f0a82/pkg/danm/danm.go#L142

in fact, it looks like you copy-pasted this exact function, rather than simply re-using the same client

pkg/danm/danm.go
```diff
@@ -183,13 +191,115 @@ func extractConnections(args *cniArgs) error {
...
+  if err != nil {
+    return "", errors.New("cannot create K8s REST client due to error:" + err.Error())
+  }
+  configmap, err := k8sClient.CoreV1().ConfigMaps(string("kube-system")).Get(string("resource-prefix-map"), meta_v1.GetOptions{})
```
Levovar: what if ConfigMap does not exist, and the parameter is not even configured?
I guess your code fails, which makes the implementation backward incompatible

the new DP support shall be optional. if parameter is not present, or ConfigMap does not exist, we should simply ignore

pkg/danm/danm.go
```diff
@@ -183,13 +191,115 @@ func extractConnections(args *cniArgs) error {
...
+  if err != nil {
+    return "", errors.New("failed to get 'resource-prefix-map' Config Map data from K8s API server due to:" + err.Error())
+  }
+  return  configmap.Data[resourceType], nil
+}
+
+func getRegisteredDevices(args *cniArgs, cniType string)([]string,error){
+  resourceMap := make(map[string]*checkpoint_types.ResourceInfo)
+  if string(args.podUid) != "" {
```
Levovar: can this ever happen? IMO Pod UUID is a mandatory param according to the CNI spec, but pls double check

if yes,remove the unnecessary complexity from the code

pkg/danm/danm.go
```diff
@@ -183,13 +191,115 @@ func extractConnections(args *cniArgs) error {
...
+    checkpoint, err := checkpoint_utils.GetCheckpoint()
+    if err != nil {
+      return nil, errors.New("failed to instantiate checkpoint object due to:" + err.Error())
+    }
+    resourceMap, err = checkpoint.GetComputeDeviceMap(string(args.podUid))
+    if err != nil {
+      return nil, errors.New("failed to retrieve Pod info from checkpoint object due to:" + err.Error())
+    }
+  }
+  prefix, err := getResourcePrefix(args, cniType)
```
Levovar: this is how this function should start: if no devices are configured, simply return and go on with network provisioning

pkg/danm/danm.go
```diff
@@ -183,13 +191,115 @@ func extractConnections(args *cniArgs) error {
...
+    return "", errors.New("cannot create K8s REST client due to error:" + err.Error())
+  }
+  configmap, err := k8sClient.CoreV1().ConfigMaps(string("kube-system")).Get(string("resource-prefix-map"), meta_v1.GetOptions{})
+  if err != nil {
+    return "", errors.New("failed to get 'resource-prefix-map' Config Map data from K8s API server due to:" + err.Error())
+  }
+  return  configmap.Data[resourceType], nil
+}
+
+func getRegisteredDevices(args *cniArgs, cniType string)([]string,error){
```
Levovar: getAllocatedDevices would be probably a better name, considering we are returning what kubelet allocated

pkg/danm/danm.go
```diff
@@ -183,13 +191,115 @@ func extractConnections(args *cniArgs) error {
...
+  if eq {
+    return nil
+  }
+  log.Printf("Required SR IOV resources: %v", requiredVfonPf)
+  log.Printf("Requested SR IOV resources: %v", requestedVfonPf)
+  return errors.New("requested and required sriov resources are not matching in the Pod definition")
+}
+
 func setupNetworking(args *cniArgs) (*current.Result, error) {
+  sriovInterfaces, interfaceDeviceMap, err := getSriovInterfaces(args)
```
Levovar: the code is not generic, and the logic is wrong. You unnecessarily increased REST traffic towards the API server by a big margin, which can lead to serious performance issues under heavy load

here is the right logic:
- you read the specific DanmNet for the specific interface inside the for loop. Once. Only once, and only the referenced. You don't read it anymore, anywhere else, e.g. inside cnidel. You re-use the same DanmNet for the rest of network provisioning
- you check if its NetworkType has any Devices configured in the ConfigMap
- if not, you do everything as right now
- if yes, you read the allocated devices with the configured name from the checkPoint file, once. You store it in a hashmap or something, for the networkType
- you choose the Xth device from the device list for this specific network provisioning operation.
- you pass it down to the network provisioner, than let it do its job. you can add it to the Interfaces struct
- you increase X for the NetworkType. You maintain X in the same hashmap, where you store the DeviceIDs

Next iteration:
- oh look, anothert sriov type DanmNet
- oh look, I already read out the allocated Devices
- oh okay, so X is at one, therefore I need to configure the second Device now for this EndPoint, and increase X

Levovar: multiple DeviceNames matching for the prefix configured for one NetworkType needs to be reconsidered though
the logic applies whatever is the case: you only read one DanmNet once for one DanmEp, you simply pass down the info to the delegating code, and you maintain an intermediate data structure to track which devices were allocated already during this operation

pkg/danm/danm.go
peterszilagyi: ConfigMap has been removed.

pkg/danm/danm.go
peterszilagyi: Code has been refactored

pkg/danm/danm.go
peterszilagyi: Code has been refactored

pkg/danm/danm.go
peterszilagyi: Implemented

pkg/danm/danm.go
peterszilagyi: Thank you for the comment. The code has been refactored accordingly. The validation part is limited to comparing the VF's PF name to the DanmNet's host_device.
I tested use-cases (not all of them) and this solution is much more permissive with deviant configurations (wrong Resource Request, missing DP, etc.). Please comment on.

pkg/danm/danm.go
```diff
@@ -183,72 +189,98 @@ func extractConnections(args *cniArgs) error {
...
-  err := syncher.GetAggregatedResult()
-  return syncher.MergeCniResults(), err
+  resourceMap, err := checkpoint.GetComputeDeviceMap(string(args.podUid))
+  if err != nil || len(resourceMap) == 0 {
+    return nil, errors.New("failed to retrieve Pod info from checkpoint object due to:" + err.Error())
+  }
+  allocatedDevices := make(map[string][]string)
+  for resourcename, resources := range resourceMap {
+    if strings.HasPrefix(resourcename, devicePool) {
+      hostDevice, err := sriov_utils.GetPfName(resources.DeviceIDs[0])
```
Levovar: code is still SR-IOV specific here
it is also a stretch to think that every networking related device in the future will be tied to a NIC
why not simply save the DeviceIDs to allocatedDevices[resourceName]?

pkg/danm/danm.go
```diff
@@ -183,72 +189,98 @@ func extractConnections(args *cniArgs) error {
...
+      if err != nil {
+        return nil, errors.New("failed to get the name of the sriov PF for device "+ resources.DeviceIDs[0] +" due to:" + err.Error())
+      }
+      allocatedDevices[hostDevice] = resources.DeviceIDs
```
Levovar: there is a one-to-one mapping between DanmNet, and devicename
once you found the device belonging to the DanmNet, you can break out from the cycle. no need to parse the rest of the checkpoint file

pkg/danm/danm.go
Levovar: better yet: just return a []string slice with the DeviceIDs belonging to one devicename

pkg/danm/danm.go
```diff
@@ -183,72 +189,98 @@ func extractConnections(args *cniArgs) error {
...
   }
-  err := syncher.GetAggregatedResult()
-  return syncher.MergeCniResults(), err
+  resourceMap, err := checkpoint.GetComputeDeviceMap(string(args.podUid))
+  if err != nil || len(resourceMap) == 0 {
+    return nil, errors.New("failed to retrieve Pod info from checkpoint object due to:" + err.Error())
+  }
+  allocatedDevices := make(map[string][]string)
+  for resourcename, resources := range resourceMap {
+    if strings.HasPrefix(resourcename, devicePool) {
```
Levovar: exact matches now, not partial anymore

pkg/danm/danm.go
```diff
@@ -183,72 +189,98 @@ func extractConnections(args *cniArgs) error {
...
-  } else {
-    cniRes, err = createDanmInterface(danmClient, iface, netInfo, args)
-    if err != nil {
-      syncher.PushResult(iface.Network, err, nil)
-      return
+    nicParams.DefaultIfaceName = "eth" + strconv.Itoa(nicID)
+    if isDelegationRequired {
+      if cnidel.IsDeviceNeeded(netInfo.Spec.NetworkType) {
+        if len(allocatedDevices) == 0 {
+          allocatedDevices, err = getAllocatedDevices(args, netInfo.Spec.Options.DevicePool)
```
Levovar: you don't need to reallocated the whole allocatedDevices map all the time. you simply need to fill up allocatedDevices[netinof.Spec.Options.DevicePool], is this record is empty (this comment is in conjuction with the other comment provided to getAllocatedDevices function

pkg/danm/danm.go
```diff
@@ -183,72 +189,98 @@ func extractConnections(args *cniArgs) error {
...
     }
-  } else {
-    cniRes, err = createDanmInterface(danmClient, iface, netInfo, args)
-    if err != nil {
-      syncher.PushResult(iface.Network, err, nil)
-      return
+    nicParams.DefaultIfaceName = "eth" + strconv.Itoa(nicID)
+    if isDelegationRequired {
+      if cnidel.IsDeviceNeeded(netInfo.Spec.NetworkType) {
+        if len(allocatedDevices) == 0 {
```
Levovar: so, if we have filled up the allocatedDevices map because of a DanmNet which had ties to DeviceA, we won't read out DeviceB of DanmNet2?

correct logic would be filling up allocatedDevices[netinfo.Spec.Options.DevicePool] if this particular record is empty

pkg/danm/danm.go
```diff
@@ -183,72 +189,98 @@ func extractConnections(args *cniArgs) error {
...
+      return cniRes, errors.New("failed to get DanmNet due to:" + err.Error())
     }
-  } else {
-    cniRes, err = createDanmInterface(danmClient, iface, netInfo, args)
-    if err != nil {
-      syncher.PushResult(iface.Network, err, nil)
-      return
+    nicParams.DefaultIfaceName = "eth" + strconv.Itoa(nicID)
+    if isDelegationRequired {
+      if cnidel.IsDeviceNeeded(netInfo.Spec.NetworkType) {
```
Levovar: combine conditions in line 237 and 238 to decrease indentation with one level

pkg/danm/danm.go
```diff
@@ -183,72 +189,98 @@ func extractConnections(args *cniArgs) error {
...
+  }
+  return allocatedDevices, nil
 }
 
-func createInterface(syncher *syncher.Syncher, iface danmtypes.Interface, args *cniArgs) {
-  danmClient, err := createDanmClient(args.stdIn)
-  if err != nil {
-    syncher.PushResult(iface.Network, err, nil)
-    return
+func getDevice(hostDevice string, allocatedDevices map[string][]string)(string, map[string][]string) {
```
Levovar: rename to something like popLastDevice, so it is more understandable what's going on here

pkg/danm/danm.go
```diff
@@ -183,72 +189,98 @@ func extractConnections(args *cniArgs) error {
...
+    nicParams.DefaultIfaceName = "eth" + strconv.Itoa(nicID)
+    if isDelegationRequired {
+      if cnidel.IsDeviceNeeded(netInfo.Spec.NetworkType) {
+        if len(allocatedDevices) == 0 {
+          allocatedDevices, err = getAllocatedDevices(args, netInfo.Spec.Options.DevicePool)
+          if err != nil {
+            log.Printf("Failed to get allocated devices due to: %v", err.Error())
+          }
+        }
+        nicParams.Device, allocatedDevices = getDevice(netInfo.Spec.Options.Device, allocatedDevices)
```
Levovar: I think it should be an error if we have ran out of Devices for a specific type

pkg/danm/danm.go
Levovar: even that validation is not needed, as it is SR-IOV specific thing that host_device, and DeviceName are related
we are getting there BTW, still some optimizations are needed (and the corrections to the smaller scope comments from the previous batch)

pkg/danm/danm.go
Levovar: you could also save one copying of the whole map if you don't return it, but rather pass it as an input parameter via pointer, and simply modify the underlying value directly within the function

pkg/danm/danm.go
peterszilagyi: Applied

pkg/danm/danm.go
peterszilagyi: Applied

pkg/danm/danm.go
peterszilagyi: Applied

pkg/danm/danm.go
peterszilagyi: Applied

pkg/danm/danm.go
peterszilagyi: Applied

pkg/danm/danm.go
peterszilagyi: Refactored

pkg/danm/danm.go
peterszilagyi: Applied

pkg/danm/danm.go
peterszilagyi: Applied

pkg/glide.yaml
```diff
@@ -20,3 +20,8 @@ import:
   version: v8.0.0
 - package: github.com/containernetworking/cni
   version: v0.6.0
+- package: github.com/intel/multus-cni
+  version: 8bf358071ab54b8f38d10027ccf81a29dfb78b3d
+- package: github.com/intel/sriov-cni
```
Levovar: you  don't need this anymore I think

pkg/danm/danm.go
```diff
@@ -183,72 +188,94 @@ func extractConnections(args *cniArgs) error {
...
-  err := syncher.GetAggregatedResult()
-  return syncher.MergeCniResults(), err
+  resourceMap, err := checkpoint.GetComputeDeviceMap(string(args.podUid))
+  if err != nil || len(resourceMap) == 0 {
+    return nil, errors.New("failed to retrieve Pod info from checkpoint object due to:" + err.Error())
+  }
+  if _, ok := resourceMap[devicePool]; !ok {
+    return nil, errors.New("failed to retrieve resources of DevicePool")
+  }
+  return &resourceMap[devicePool].DeviceIDs, nil
```
Levovar: a slice is actually a pointer anyway, so it is kind of strange to dereference it

what I meant is passing the map by a pointer

I think the most elegant way would be creating some named types, like Devices []string, and DeviceMap map[string]Devices
then pass *DeviceMap to popDevice as input

would increase readability IMO by a big margin, but I leave it up to you

pkg/cnidel/cniconfs.go
```diff
@@ -45,25 +45,18 @@ func readCniConfigFile(netInfo *danmtypes.DanmNet) ([]byte, error) {
...
-    Type:   "sriov",
-    PfName: netInfo.Spec.Options.Device,
-    IfName: ep.Spec.Iface.Name,
-    L2Mode: true,
-    Vlan:   vlanid,
-    Dpdk:   DpdkOption{},
-    Ipam:   ipamOptions,
+    Name:      netInfo.Spec.NetworkID,
+    Type:      "sriov",
+    PfName:    netInfo.Spec.Options.Device,
```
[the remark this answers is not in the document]
Levovar: I think in the new version there is anyway only Master: 
https://github.com/intel/sriov-cni/blob/master/pkg/types/types.go#L22

we have agreed to integrate from commit hash. right?

Levovar:
Okay, logic is now super-duper.
One cosmetical, readability related comment left, and the ones from the first batch related to making the new struct parameters generic, deleting unneeded parameters from the schema, nad correcting documentation

let's also find out why netwatcher ignores your new param. you know that you use device_pool as the name of your attribute in your code, right? weren't you using "resource_pool" in your test examples?
that would explain why netwatcher can't parse it :)

pkg/crd/apis/danm/v1/types.go
```diff
@@ -97,6 +97,7 @@ type DanmEpIface struct {
   MacAddress  string            `json:"MacAddress"`
   Proutes     map[string]string `json:"proutes"`
   Proutes6    map[string]string `json:"proutes6"`
+  VfDeviceID  string            `json:"VfDeviceID,omitempty"`
```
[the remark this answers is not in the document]
peterszilagyi: Applied

pkg/cnidel/types.go
```diff
@@ -29,17 +31,36 @@ type sriovNet struct {
...
+  // CNI binary location
+  CNIDir string `json:"cniDir"`
+  // Device PCI ID
+  DeviceID string `json:"deviceID"`
+  // Device Info
+  DeviceInfo *VfInformation `json:"deviceinfo,omitempty"`
 }
 
-// DpdkOption represents the DPDK options for the sriov plugin
+// VfInformation is a DeviceIfo desctiprtor expected by sriov plugin v1.0.0
```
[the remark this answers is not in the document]
peterszilagyi: Applied

pkg/cnidel/types.go
```diff
@@ -29,17 +31,36 @@ type sriovNet struct {
   // IPAM configuration to be used for this network
   Ipam   danmtypes.IpamConfig `json:"ipam,omitEmpty"`
   // DPDK configuration
-  Dpdk   DpdkOption `json:"dpdk,omitEmpty"`
+  Dpdk   *DpdkOption `json:"dpdk,omitEmpty"`
```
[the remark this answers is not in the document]
peterszilagyi: Applied

pkg/cnidel/cniconfs.go
```diff
@@ -45,25 +45,18 @@ func readCniConfigFile(netInfo *danmtypes.DanmNet) ([]byte, error) {
...
+    PfBackward: netInfo.Spec.Options.Device,
```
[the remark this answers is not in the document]
peterszilagyi: Applied

pkg/cnidel/cniconfs.go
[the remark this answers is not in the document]
peterszilagyi: Fixed

pkg/glide.yaml
peterszilagyi: Removed

peterszilagyi:
Thanks for comments. Code related stuffs have been corrected. README.md update is going to be happened soon.

> let's also find out why netwatcher ignores your new param. you know that you use device_pool as the name of your attribute in your code, right? weren't you using "resource_pool" in your test examples?
> that would explain why netwatcher can't parse it :)

I'm gonna doublecheck everything. BTW: I rebuild and redeployed all component after the last rebase from upstream code. So all component have to have the same understanding about the scheme.
FYI: while I was testing with in-printed device_pool name i was realizing that the given DanmNet's device_pool attribute had been updated properly right after the Pod creation. The assumption is because of IPAM updates the DanmNet it triggers a re-validation and at that time netwatcher applies the proper scheme... This is what I will investigate today.

pkg/cnidel/cniconfs.go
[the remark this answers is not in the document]
Levovar: ohmm, I'm not sure this will be okay. IMO CNI expects the name of the physical device from the host, not the logical resource name from Kubernetes
my comment was simply that we can figure out pretty easily the name of the PF based on the DeviceID, right? If yes, let's just do that, instead of expecting the network administrator to know which host_device the Kubernetes belongs maps to. that anyway only works as long as the mapping is 1:1

pkg/cnidel/cniconfs.go
```diff
@@ -45,25 +47,17 @@ func readCniConfigFile(netInfo *danmtypes.DanmNet) ([]byte, error) {
...
-    Vlan:   vlanid,
-    Dpdk:   DpdkOption{},
-    Ipam:   ipamOptions,
+    Name:      netInfo.Spec.NetworkID,
+    Type:      "sriov",
+    PfName:    netInfo.Spec.Options.Device,
+    L2Mode:    true,
+    Vlan:      vlanid,
+    Ipam:      ipamOptions,
+    DeviceID:  ep.Spec.Iface.DeviceID,
```
Levovar: also, will this work without filling-in DeviceInfo struct too?
I mean I guess technically it will work because of this code: https://github.com/intel/sriov-cni/blob/master/sriov/main.go#L38, but it won't pickup the exact Device we have wanted, it will allocate one for itself from the same PF, but that's not what we want

Levovar: or maybe this is the intended behaviour from our side? :)
I will sleep on it during the weekend

peterszilagyi: According to our tests, it works as expected. The CNI picks up the interface specified in DeviceID.
check this please: https://github.com/intel/sriov-cni/blob/c97db95e4355fa247116608dcdb0da2f18f02140/pkg/config/config.go#L45

Levovar: yes you are right!
perfect

pkg/cnidel/cniconfs.go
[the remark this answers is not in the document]
peterszilagyi: Thanks. Most probably you are right so I modified the code accordingly.
BTW: this was the original implementation

README.md
```diff
@@ -343,6 +345,130 @@ The CNI provisions IPVLAN interfaces in L2 mode, and supports the following extr
 * provisioning generic IP routes into a configured routing table inside the Pod's network namespace
 * Pod-level controlled provisioning of policy-based IP routes into Pod's network namespace
 
+#### Integrating SRIOV CNI
+
+DANM has got interface to the latest Intel SR-IOV CNI. The solution relies on the Intel SR-IOV Network Device Plugin which maintains the actual resource allocation in kubernetes checkpoint. DANM utilizes the checkpoint data and passes the PCI ID of VF to be attached to the Pod to SR-IOV CNI.
```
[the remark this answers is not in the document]
peterszilagyi: invalid content. has been removed.

README.md
```diff
@@ -343,6 +345,130 @@ The CNI provisions IPVLAN interfaces in L2 mode, and supports the following extr
...
+		apiVersion: v1
+		metadata:
+			name: resource-prefix-map
+			namespace: kube-system
+		kind: ConfigMap
+		data:
+			sriov: nokia.k8s.io/sriov
+			foo: foobaa.com/baa
+  ```
+  This Config Map contains resource type and resource name prefix bindings. DANM looks up for the value of the "sriov" key and looking for matching Pod resource in the checkpoint file where the name starts with this value.
```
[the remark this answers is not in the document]
peterszilagyi: invalid content. has been removed.

README.md
```diff
@@ -343,6 +345,130 @@ The CNI provisions IPVLAN interfaces in L2 mode, and supports the following extr
...
+			- resourceName: "sriov_ens1f1"
+			  rootDevices: [ "03:00.1" ]
+			  sriovMode: true
+			  deviceType: netdevice
+  ```
+* **In case of RBAC, DANM requires getter role for configmaps**
+
+#### Using SRIOV CNI
+
+* Create sriov type DanmNets on the usual way. **Do not use** container_prefix option, SR-IOV CNI manages the interface naming automatically.
```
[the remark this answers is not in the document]
peterszilagyi: invalid content. has been removed.

README.md
```diff
@@ -343,6 +345,130 @@ The CNI provisions IPVLAN interfaces in L2 mode, and supports the following extr
...
+```
+Example:
```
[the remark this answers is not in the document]
peterszilagyi: invalid content. has been removed.

README.md
```diff
@@ -343,6 +345,130 @@ The CNI provisions IPVLAN interfaces in L2 mode, and supports the following extr
...
+Example: 
+	apiVersion: danm.k8s.io/v1
+	kind: DanmNet
+	metadata:
+	  name: sriov-danm1
+	spec:
+	  NetworkID: sriov-danm1
+	  NetworkType: sriov
+	  Options:
+		host_device: ens1f0
```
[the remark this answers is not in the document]
peterszilagyi: invalid content. has been removed.

pkg/cnidel/cniconfs.go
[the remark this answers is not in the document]
Levovar: yeah it actually works - as long as you, the developer, who know this stuff intimately- configures it.
once Average Joe starts using your feature, and as his first step misconfigures the name of the PF in DanmNet, that's when trouble starts :)

better to limit the possibilities of user misconfiguration, less fault tickets for us.
in case of NetworkType: sriov we will simply ignore Device from now on

pkg/danm/danm.go
peterszilagyi: If you would accept, I would leave it as it is

README.md
```diff
@@ -343,6 +345,130 @@ The CNI provisions IPVLAN interfaces in L2 mode, and supports the following extr
...
+		  requests:
+			nokia.k8s.io/sriov_ens1f0: '3'
+			nokia.k8s.io/sriov_ens1f1: '3'
+		  limits:
+			nokia.k8s.io/sriov_ens1f0: '3'
+			nokia.k8s.io/sriov_ens1f1: '3'
+	  nodeSelector:
+		nodename: caas_worker1
+```
+**ATTENTION: The total amount of requested sriov interfaces must be equal to the total amount of sriov resource requests. In case of any mismatch, DANM blocks the Pod instantiation and reports the deviance in /var/log/plugin.log**
```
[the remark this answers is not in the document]
peterszilagyi: invalid content. has been removed.

pkg/danm/danm.go
Levovar: yeah no worries, maybe I will re-factor on my own when I'm bored

schema/DanmNet.yaml
```diff
@@ -36,6 +36,12 @@ spec:
     # A Virtual Function belonging to this Physical Function is taken-up in case NetworkType is set to SRIOV.
     # OPTIONAL - STRING
     host_device: ## PARENT_DEVICE_NAME ##
+    # Name of an SR-IOV Network Device Plugin resource pool
```
Levovar: this doc also needs to be reworked to talk about device management in general, not just exclusively from SR-IOV VFs
SR-IOV is just one example of using Kubernetes devices during network provisioning

schema/network_attach.yaml
```diff
@@ -65,6 +65,15 @@ spec:
...
           ]
-
   ...
-
+  spec:
+    containers:
+    - name:
+    ...
+      resources:
+      # In case of sriov interface type, resources must be requested from the Intel SR-IOV Netwrok Device plugin
```
Levovar: same. any devices related to network management can be requested via this API, not just SR-IOV VFs

schema/DanmNet.yaml
peterszilagyi: Sorry, I have no idea what kind of other device plugin can come into the picture from networking viewpoint. I've found this list but have no clue how to refer them (except sriov) with regards DANM.
https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/#examples 
On the other hand my opinion is that the device_pool attribute description should have SR-IOV specific part due to danm has specific assumptions. (eg.: one-to-one mapping)

Levovar: nope.
we are only assuming one device pool is mapped to one DanmNet. but that is the whole idea behind the generic code I made you implement:
- multiple DanmNets can refer the same Device Pool
- one Device pool can contain pretty much any number of devices, related to multiple physical "stuff"

that is not 1:1 mapping!

The device_pool parameter _generally_ represents the base resource name of the Kubernetes Devices connected to this network.
_Specifically_ for DanmNets with NetworkType: sriov this option is mandatory, and it represents a pool of VFs Pods connecting to this network will get on provisioned.

You see the difference, I hope. Just because we only have one supported use-case of a feature _right now_ , the feature, and the interfaces themselves shall be still generic, and easily re-usable  in the future.
Hence you work on this review for a week now.

Hope it was a good lesson for future endeavours :)

schema/DanmNet.yaml
```diff
@@ -36,6 +36,12 @@ spec:
     # A Virtual Function belonging to this Physical Function is taken-up in case NetworkType is set to SRIOV.
     # OPTIONAL - STRING
     host_device: ## PARENT_DEVICE_NAME ##
+    # Name of a network Device Plugin resource pool
+    # The device_pool parameter generally represents the base resource name of the Kubernetes Devices connected to this network.
+    # Specifically for DanmNets with "NetworkType: sriov" this option is mandatory, and it represents a pool of VFs Pods connecting to this network will get one provisioned from.
+    # container_prefix must be ommited, Intel SR-IOV CNI generates unique interface name automatically
```
Levovar: I commented at the beginning of this PR too, that container_prefix option is working with SR-IOV CNI, and it is DANM which generates the unique
host_device is the parameter which is ignored from now on in a NetworkType: sriov DanmNet
Ignorance of a specific attribute should be documented in the description of the attribute itself

peterszilagyi: Applied.
Thanks. This line remaied there due to editing failure, the intence was to remove it.

Levovar:
So, all-in-all, thanks! I think it was a good journey :)
I'm fine with README and UTs arriving separately, let's not prolong this review anymore

peterszilagyi:
I agree :) Thank you for driving me to reach this point.

PR #53, opened 2019-03-06
add unit TC to PopDevice function

tamaskopcso:
**What type of PR is this?**

failing-test

Unit Test and increases code coverage

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/danm/danm_test.go
+++ b/pkg/danm/danm_test.go
@@ -7,23 +7,18 @@ import (
 var devicePool0 = "pool0"
 var devicePool1 = "pool1"
 
-var allocatedDevices = make(map[string]*[]string)
+func TestPopDevice(t *testing.T) {
+  allocatedDevices := make(map[string]*[]string)
 
-func TestPopDevicePanic(t *testing.T) {
-    defer func() {
-        if r := recover(); r == nil {
-            t.Errorf("Uninitialized map should expect error.")
-        }
-    }()
+  device,err := popDevice(devicePool0, allocatedDevices)
+  if err == nil {
+    t.Errorf("Uninitialized map should expect error.")
+  }
 
-    popDevice(devicePool0, allocatedDevices)
-}
-
-func TestPopDevice(t *testing.T) {  
   allocatedDevices[devicePool0] = &[]string{"device0"}
   allocatedDevices[devicePool1] = &[]string{"device1", "device2"}
-  
-  device,err := popDevice(devicePool1, allocatedDevices)
+
+  device,err = popDevice(devicePool1, allocatedDevices)
   if device != "device2"  || err != nil {
     t.Errorf("Received device or error does not match with expectation.")
   }
--- a/pkg/danm/danm_test.go
+++ b/pkg/danm/danm_test.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+  "testing"
+)
+
+var devicePool0 = "pool0"
+var devicePool1 = "pool1"
+
+var allocatedDevices = make(map[string]*[]string)
+
+func TestPopDevicePanic(t *testing.T) {
+    defer func() {
+        if r := recover(); r == nil {
+            t.Errorf("Uninitialized map should expect error.")
+        }
+    }()
+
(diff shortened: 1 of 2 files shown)


Levovar:
wow, the system works! even better than I expected :D

actually I was wondering yesterday why cniconfs hasn't shown-up in the coverage detail. now that you added a UT touching it, the whole file shows up finally in the project's overall coverage status, and since you haven't covered the full file 100%, overall coverage decreased

cool. I wasn't even aware coverall fails the check in this case

pkg/danm/danm_test.go
```diff
@@ -0,0 +1,38 @@
+package main
```
Levovar: yeah, you can't test this in other way, but just as an FYI I will detail how this should be done properly, in pratice, for future reference

So, in Golang, you can't really test the main function. also testing the main package, containing the main function, is not really preferred / good / solved / allowed. this is also why the overall coverage of the whole project hasn't increased, because you are not actually testing a package
unfortunately for you, the code is currently structured in a wrong way. more standard way would be creating a cmd directory , and relocating the "main" functions of all binaries there. and refactor the main functions, that they only contain a couple of LOCs before calling something from ...pkg/..., which is no testable

I'm gonna do this later, so no worries

pkg/danm/danm_test.go
```diff
@@ -0,0 +1,38 @@
...
+    defer func() {
+        if r := recover(); r == nil {
+            t.Errorf("Uninitialized map should expect error.")
+        }
+    }()
+
+    popDevice(devicePool0, allocatedDevices)
+}
+
+func TestPopDevice(t *testing.T) {
```
Levovar: however this unit testing paradigm is highly advised against, and very much not preferred in this project (I know specifically within Golang people do this, that doesn't mean much to me though :) )
packages should be tested through their external interfaces, not directly through their internals. this is why all the existing test files are actually located in a totally different package

pkg/danm/danm_test.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+  "testing"
+)
+
+var devicePool0 = "pool0"
+var devicePool1 = "pool1"
+
+var allocatedDevices = make(map[string]*[]string)
+
+func TestPopDevicePanic(t *testing.T) {
```
Levovar: this test case is one of the most common antip-atterns of unit testing

if you find an error within the production code with your unit test, you should correct the error in the code, rather than glorifying it with a test case
in this case popDevice shall return an error to the caller if there is nothing to pop

pkg/danm/danm_test.go
Levovar: which might be hard to do within this specific code, and this is exactly the reason why I haven't started writing UTs for this specific file yet. needs more thorough consideration

we can live with this for the time being, again just an FYI for you regarding best practices

pkg/danm/danm_test.go
Levovar: FYI: this is not an FYI comment like the other two, this is malpractice

on the other hand: congrats, you found an error in the code! :)

pkg/danm/danm_test.go
Levovar: for reference see the IPAM test package, and how I only test it through its Allocate, and Free calls.
this is exactly like black-box testing, just in a smaller scope

TothFerenc:
retest

pkg/danm/danm_test.go
```diff
@@ -0,0 +1,33 @@
...
+  device,err = popDevice(devicePool1, allocatedDevices)
+  if device != "device2"  || err != nil {
+    t.Errorf("Received device or error does not match with expectation.")
+  }
+  device,err = popDevice(devicePool1, allocatedDevices)
+  if device != "device1"  || err != nil {
+    t.Errorf("Received device or error does not match with expectation.")
+  }
+  device,err = popDevice(devicePool1, allocatedDevices)
+  if device == ""  && err == nil {
```
Levovar: in erroneous cases it is kind of standard to not expect anything regarding the value of the "other" return parameter
checking for error is enough

PR #60, opened 2019-03-14
100% coverage for ipam package - and bugfixes

Levovar:
Bugs caught and corrected during UTing this package:
1: Last IP from the allocation pool was never used
2: When Pool.End was not provided, we set the first IP into this parameter, rather than the last.
This has resulted in failing validation for a subnet providing only CIDR without allocation pool
3: Static IPs were not validated against the allocation pool
4: When the allocation pool is entirely exhausted, we did not explicitly return an error, but let DANM proceed with setting up an interface without an IP. This has been changed now

### DIFF
--- a/pkg/ipam/ipam.go
+++ b/pkg/ipam/ipam.go
@@ -146,7 +146,7 @@ func allocIPv4(reqType string, netInfo *danmtypes.DanmNet, ip4 *string) (error) 
   } else {
     ip, ipnet, _ := net.ParseCIDR(reqType)
     if ip == nil {
-      return errors.New("IPv4 allocation failure, invalid static IP requested!")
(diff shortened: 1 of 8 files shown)


Levovar:
5: Added a new check, explicitly returning an error if an IPv6 address is requested for a subnet smaller than /64
6: Asking for an IP from a L2 IPv6 network now properly results in an explicit error returned to the caller, in both static, nad dynamic case
7: Requesting invalid, or malformed static IPv6 addresses now results in an explicit error returned to the caller

Levovar:
8: if someone would have changed the allocated IP in a DanmEp to an invalid, or out-of-Cidr-range IP, ipam.Free wold have crashed and paniced
Added some checks to protect against such network /user corruption cases

PR #69, opened 2019-03-28
Set IPv6 related sysctls, if at least one IPv6 CIDR is set on any interfaces

### LINKED ISSUE #68: IPv6 does not work because IPv6 is disabled in Pod's network namespace

feature request in a sense that IPv6 sysctl support is missing,
but it can be a bug, because IPv6 support is broken

**What happened**:

CNIs fail to set IPv6 address(es) on interfaces, because IPv6 is disabled (sysctl: `disable_ipv6=1`) in the Pod's network namespace.

IPv6 should work as desired.

**How to reproduce it**:

Simple DanmNet+Pod definition which requests IPv6 on at least one interface.

DANM should take care of IPv6 related sysctls before CNIs are invoked.

TothFerenc:
**What type of PR is this?**

feature

DANM takes care of IPv6 related sysctls before CNIs are invoked.

Fixes #68 

Tested with ipvlan inside VMs.

None.

### DIFF
--- a/pkg/metacni/metacni.go
+++ b/pkg/metacni/metacni.go
@@ -6,6 +6,7 @@ import (
   "log"
   "net"
   "os"
+  "runtime"
   "strconv"
   "strings"
   "encoding/json"
@@ -13,6 +14,8 @@ import (
   "github.com/containernetworking/cni/pkg/skel"
   "github.com/containernetworking/cni/pkg/types"
   "github.com/containernetworking/cni/pkg/types/current"
+  "github.com/containernetworking/plugins/pkg/ns"
+  "github.com/containernetworking/plugins/pkg/utils/sysctl"
   meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
   k8s "k8s.io/apimachinery/pkg/types"
   "k8s.io/client-go/rest"
@@ -207,14 +210,46 @@ func popDevice(devicePool string, allocatedDevices map[string]*[]string)(string,
   return device, nil
 }
 
+func preparePodForIpv6(args *cniArgs) error {
+  runtime.LockOSThread()
+  defer runtime.UnlockOSThread()
+  // save the current namespace
+  origNs, err := ns.GetCurrentNS()
+  if err != nil {
+    return errors.New("failed to get current network namespace due to:" + err.Error())
+  }
+  // enter to the Pod's network namespace
+  podNs, err := ns.GetNS(args.netns)
+  if err != nil {
+    return errors.New("failed to get Pod's network namespace due to:" + err.Error())
+  }
+  err = podNs.Set()
+  if err != nil {
+    return errors.New("failed to switch to Pod's network namespace due to:" + err.Error())
+  }
+  defer func() {
+    podNs.Close()
+    origNs.Set()
+  }()
+  // set sysctl
+  _, err = sysctl.Sysctl("net.ipv6.conf.all.disable_ipv6", "0")
+  if err != nil {
+    return errors.New("failed to set sysctl due to:" + err.Error())
+  }
+  return nil
+}
+
 func setupNetworking(args *cniArgs) (*current.Result, error) {
   danmClient, err := createDanmClient(args.stdIn)
   if err != nil {
     return nil, errors.New("failed to create DanmClient due to:" + err.Error())
   }
+  err = preparePodForIpv6(args)
+  if err != nil {
+    return nil, errors.New("failed to prepare Pod for IPv6 due to:" + err.Error())
+  }
   syncher := syncher.NewSyncher(len(args.interfaces))
   allocatedDevices := make(map[string]*[]string)
-
   checkpoint, err := checkpoint_utils.GetCheckpoint()
   if err != nil {
     return nil, errors.New("failed to instantiate checkpoint object due to:" + err.Error())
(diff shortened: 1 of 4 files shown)


pkg/metacni/metacni.go
```diff
@@ -207,14 +210,54 @@ func popDevice(devicePool string, allocatedDevices map[string]*[]string)(string,
...
+}
+
 func setupNetworking(args *cniArgs) (*current.Result, error) {
   danmClient, err := createDanmClient(args.stdIn)
   if err != nil {
     return nil, errors.New("failed to create DanmClient due to:" + err.Error())
   }
+  // set IPv6 related sysctls, if at least one IPv6 CIDR is set on any interfaces
+  for _, nicParams := range args.interfaces {
+    if nicParams.Ip6 != "" {
```
Levovar: just because somebody asked for IPv6 from Pod Spec, it doesn't automatically mean the network he is connecting to is also IPv6 enabled
presence of Net6 should be the indicator we need to make this setting

pkg/metacni/metacni.go
```diff
@@ -207,14 +210,54 @@ func popDevice(devicePool string, allocatedDevices map[string]*[]string)(string,
...
+  if err != nil {
+    return errors.New("failed to switch to Pod's network namespace due to:" + err.Error())
+  }
+  defer func() {
+    podNs.Close()
+    origNs.Set()
+  }()
+  // set sysctls
+  log.Println("IPv6 sysctl: net.ipv6.conf.all.disable_ipv6 = 0")
+  _, err = sysctl.Sysctl("net.ipv6.conf.all.disable_ipv6", "0")
```
Levovar: I would prefer per interface setting, rather than a generic enable it over all

pkg/metacni/metacni.go
Levovar: BTW we are already looping over the interfaces map a couple of lines later.
combining with the other comment asking for per interface  setting, you can just put this could into the scope of the existing loop

pkg/metacni/metacni.go
```diff
@@ -207,14 +210,54 @@ func popDevice(devicePool string, allocatedDevices map[string]*[]string)(string,
...
+  err = podNs.Set()
+  if err != nil {
+    return errors.New("failed to switch to Pod's network namespace due to:" + err.Error())
+  }
+  defer func() {
+    podNs.Close()
+    origNs.Set()
+  }()
+  // set sysctls
+  log.Println("IPv6 sysctl: net.ipv6.conf.all.disable_ipv6 = 0")
```
Levovar: don't do unconditional, unnecessary logs

pkg/metacni/metacni.go
```diff
@@ -207,14 +210,54 @@ func popDevice(devicePool string, allocatedDevices map[string]*[]string)(string,
   return device, nil
 }
 
+func setIpv6Sysctls(args *cniArgs) error {
```
Levovar: can we make this generic?

what I'm thinking is that we anyway need to set a couple of kernel level parameters for multiple reasons in the future, probably. just for IPv6 we already know we need to set some others later on
also setting the parameter is orthogonal to the backend used

how about adding a new list to the DanmEp, something like "kernel_settings"?
and during putDanmEp call we could simply transparently set the full content of this list into the Pod netns

pkg/metacni/metacni.go
```diff
@@ -207,14 +210,54 @@ func popDevice(devicePool string, allocatedDevices map[string]*[]string)(string,
...
+  runtime.LockOSThread()
```
Levovar: regardless of the implementation specifics: Pod NetNS manipulation code we usually put into danmep package, and invoke through an interface
I think this kind of code would also belong there (possibly together with the function "putDanmEp", if you decide to implement my comment

pkg/metacni/metacni.go
TothFerenc: code has been refactored heavily...

pkg/metacni/metacni.go
TothFerenc: Done, but using invert logic.
We enable IPv6 on all interfaces (in order to run CNI smoothly), then we disable IPv6 where it is not needed.

pkg/metacni/metacni.go
TothFerenc: code has been refactored heavily...

pkg/metacni/metacni.go
TothFerenc: code has been refactored heavily...

pkg/metacni/metacni.go
TothFerenc: I will remove such lines later...

pkg/metacni/metacni.go
TothFerenc: Done.

TothFerenc:
The code has been refactored heavily. Please review it again, thanks!

pkg/metacni/metacni.go
```diff
@@ -269,6 +304,10 @@ func createDelegatedInterface(syncher *syncher.Syncher, danmClient danmclientset
   if err != nil {
     syncher.PushResult(iface.Network, errors.New("DanmEp object could not be PUT to K8s due to error:" + err.Error()), nil)
   }
+  err = danmep.SetDanmEpSysctls(ep)
+  if err != nil {
+    syncher.PushResult(iface.Network, errors.New("Sysctls could not be set due to error:" + err.Error()), nil)
```
Levovar: it might make me silly, but this code made me thinking... shouldn't we immediately return after every failed syncher.PushResult calls? :)

Levovar:
Looks cool!

Except that I just now realized that we should have returned from an asynch thread after the first error all this time :)

pkg/metacni/metacni.go
TothFerenc: Done

Levovar: haven't you forgot to git push?

PR #76, opened 2019-04-15
More sysctls to set for IPv6

TothFerenc:
**What type of PR is this?**
feature

More sysctls need to be set for IPv6.

Enhances #68 

NONE

### DIFF
--- a/pkg/danmep/danmep.go
+++ b/pkg/danmep/danmep.go
@@ -18,16 +18,30 @@ const (
   dockerApiVersion = "1.22"
 )
 
-type sysctlData struct {
+type sysctlFunction func(danmtypes.DanmEp) bool
+type sysctlObject struct {
   sysctlName  string
   sysctlValue string
 }
-var sysctls = map[string][]sysctlData {
-  "enable_ipv6_on_iface": {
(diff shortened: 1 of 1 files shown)


pkg/danmep/danmep.go
```diff
@@ -139,8 +141,16 @@ func SetDanmEpSysctls(ep danmtypes.DanmEp) error {
     podNs.Close()
     origNs.Set()
   }()
-  // set sysctls for IPv6 (since IPv6 is enabled on all interfaces, let's disable on those where it is not needed)
-  if ep.Spec.Iface.AddressIPv6 == "" {
+  // set sysctls on those interfaces where IPv6 is needed
+  if ep.Spec.Iface.AddressIPv6 != "" {
```
Levovar: this looks like now a lot of specific code, which is not easiyl extendable

how about a little rework like this:
- we maintain the following data structure: a function reference returning a bool, and a list of stuff to set
- the sysctl setting code iterates through the sysctl records, call the referenced function, if that return true we blindly set what is in the list

Levovar:
Generally I'm not against multiple commits if they make sense, but I despise merge commits
Could you kindly squash them into the functional change, or simply rebase your functional commit on top of origin/master HEAD?

PR #79, opened 2019-04-18
Replace external arping binary execution with Go implementation

### LINKED ISSUE #75: gARP is not needed for IPv6
bug

**What happened**:
Got warning message in plugin.log:
```
2019/04/15 15:45:43 WARNING: sending gARP Reply failed with error:gARP update for IP address: fc00:caa5:1:a:f816:3eff:fe63:a6c0 was unsuccessful:exit status 2  , but we will ignore that for now!
```

Avoid unnecessary warning messages.

**How to reproduce it**:
Request IPv6 address on ipvlan interface.

TothFerenc:
**What type of PR is this?**
fix

fix IPv6 gARP issue, cleaner code

Fixes #75 

Now it produces valid gARP packet, but I keep it as WIP until I can test it on a bare metal system where I have access to the switches. 

NONE

### DIFF
--- a/pkg/danmep/ep.go
+++ b/pkg/danmep/ep.go
@@ -7,11 +7,10 @@ import (
   "os"
   "runtime"
   "strconv"
-  "strings"
-  "os/exec"
   "github.com/vishvananda/netlink"
   "github.com/containernetworking/plugins/pkg/ns"
   danmtypes "github.com/nokia/danm/crd/apis/danm/v1"
+  "github.com/j-keck/arping"
 )
 
 func createIpvlanInterface(dnet *danmtypes.DanmNet, ep danmtypes.DanmEp) error {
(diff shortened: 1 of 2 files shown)


pkg/danmep/ep.go
```diff
@@ -118,42 +121,19 @@ func createContainerIface(ep danmtypes.DanmEp, dnet *danmtypes.DanmNet, device s
   if err != nil {
     return errors.New("cannot set renamed IPVLAN interface to up because:" + err.Error())
   }
-  sendGratArps(ip, ip6, dstPrefix)
+  if ip != "" {
+    err = arping.GratuitousArpOverIfaceByName(addr4, dstPrefix)
+    if err != nil {
+      return errors.New("cannot send gARP because:" + err.Error())
```
Levovar: let's not promote this event from warning to error which in turn will kill the whole Pod
otherwise looking good

TothFerenc: done

TothFerenc:
Tested it on bare metal environment: using ipvlan danmnets right after Pod creation MAC addresses are appeared in the VLAN specific MAC address table on the leaf switches. 
Ready to merge.

PR #98, opened 2019-06-21
Delegate CNIs sometimes do not provision the IPs we asked them

Levovar:
In this case we will throw an error from now on. This will ultimately result in a retry, which might solve the issue if it was only a temporary one.


Levovar:
closing, as the issue is not relevant anymore, and the reported symptom had another root cause

PR #134, opened 2019-08-22
Set verbosity level on info/debug messages to reduce unnecessary logging

### LINKED ISSUE #14: Svcwatcher floods std:out even with default log level

TothFerenc:
**What type of PR is this?**
cleanup

Currently svcwatcher is very noisy. It generates too many log entries, because all the function calls are logged. This PR sets the verbosity level of these info/debug messages to 5, so they will be suppressed by default. Such detailed logging can be enabled by `-v=5` commandline parameter.

**Which issue(s) this PR fixes**
Fixes #14 

I know that you have plans to replace glog to something better, but this fix is kindof urgent to reduce fluentd load and save log data store capacity.

NONE


Levo-var:
Yeah replacement can be done later, sure
LGTM but I can't merge from my holiday account :)

PR #155, opened 2019-09-27
Adding support for Intel Fortville + DPDK use-case

Levovar:
PR solves bullet point 2 described in https://github.com/nokia/danm/issues/148.

It also makes sure to skip kernel tunings for interfaces which are actually not managed by the kernel.


Levovar:
Actually...
This PR is now a generalization of how we do "post-processing" for our interfaces, including setting kernel configs, and creating IP routes etc.
The code is reworked to be entirely generic for all links, under all circumstances.

This greatly streamlines execution paths, and allows to finally get rid of our own IPVLAN interface handling module. All special features originally working only for IPVLAN are now 100% work for everything - even for VFIO bound PCI devices!

PR #169, opened 2019-12-27
PUT danmep before invoking delegated plugins

### LINKED ISSUE #165: IP leak for the delegated CNI plugin (Calico we use) when there are time-out issues

>
 bug
> feature

**What happened**:
As a background, there is some kind of performance issue in the cluster.
When DANM invoked Calico to add interface to pod, Calico did not response in time (10s), but it did run the add process for the pod. After the waiting time-out, DANM did not create danmep for this calico interface and returned error. Then when dealing with the DEL operation, DANM will not invoke Calico to delete because there is no danmep about the interface, as a result, IP leak in Calico.

A danmep should be created for the delegated CNI plugin even if the invoking time out, so an explicit DEL call can be performed when deleting the pod.

**How to reproduce it**:

**Environment**:
- DANM version (use `danm -version`):
- Kubernetes version (use `kubectl version`):
- DANM configuration (K8s manifests, kubeconfig files, CNI config file):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Others:

clivez:
**What type of PR is this?**
>
 bug
> cleanup
> design
> documentation
> failing-test
> feature

When creating delegated interface, put the danmep into etcd before invoking the delegated plugin, and update the danmep after the invoking finished successfully. So a danmep for this interface can be stored in the etcd even if the invoking is time out.
 
Fixes #165 

None

None

### DIFF
--- a/pkg/metacni/metacni.go
+++ b/pkg/metacni/metacni.go
@@ -367,6 +367,12 @@ func createDelegatedInterface(syncher *syncher.Syncher, danmClient danmclientset
     syncher.PushResult(netInfo.ObjectMeta.Name, errors.New("DanmEp object could not be created due to error:" + err.Error()), nil)
     return
   }
+  err = danmep.PutDanmEp(danmClient, ep)
+  if err != nil {
+    cnidel.FreeDelegatedIps(danmClient, netInfo, ep.Spec.Iface.Address, ep.Spec.Iface.AddressIPv6)
+    syncher.PushResult(netInfo.ObjectMeta.Name, errors.New("DanmEp object could not be PUT to K8s due to error:" + err.Error()), nil)
+    return
+  }
   delegatedResult,err := cnidel.DelegateInterfaceSetup(DanmConfig, danmClient, netInfo, &ep)
   if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/metacni/metacni.go
```diff
@@ -367,6 +367,12 @@ func createDelegatedInterface(syncher *syncher.Syncher, danmClient danmclientset
     syncher.PushResult(netInfo.ObjectMeta.Name, errors.New("DanmEp object could not be created due to error:" + err.Error()), nil)
     return
   }
+  err = danmep.PutDanmEp(danmClient, ep)
```
Levovar: yeah I was thinking about how to approach this, and yes this is one way to do it :)

my issue with this approach is that we add an extra REST API call for ALL successful transactions, because SOME of them might fail. albeit not much, but it is extra latency

instead, I was thinking about leaving the existing DanmEp creation as-is. Instead I would simply add a PutDanmEp call into the error branch DelegateInterfaceSetup.
Same effect, but without the extra leg for the successful calls - which are the majority

Levovar:
abandoned by author, will be picked up in https://github.com/nokia/danm/pull/188

Issue #147, opened 2019-09-17
Support to create ipv6 danm network with different netmask


ArunN85:
As of now in Danm3, pure ipv6 support is not present and also not able to create network with mask as /120.
Only /64 is allowed as of now Is this support available in DANM4. Also this danm network when created with /64 and /120, what all will be the changes with respect to the gateway for the POD's

Levovar:
the correct statement is that up until /64 is allowed, not specifically only /64
/120 means 256 addresses. IPv4 is perfectly suitable IMO for such a network. can you elaborate on your use-case, why would you like to use IPv6 for such a small pool?

Levovar:
BTW full IPv6 support was introduced with DANM 3.3

ArunN85:
I wanted to plan the ipv6 network with /120 netmask and in some customer places these are used.
If my network plan has /120 as the netmask, then it will not be possible to define it. These kind of small subnets are used when there are multiple LANS needs to be configured for my telco network,

Which ip will danm ip consider as the gateway if it is /64 and /120 ?

Levovar:
just to be clear: you state there are customer environments where IPv4 is entirely not allowed, but all you get is a /120 IPv6 subnet?

DANM doesn't overwrite the default GW, it just creates the IP, rest is up to kernel. If you want to set up IP routes you can do that via Spec.Options.Routes6, where you supply the GW IP

the reason we don't support smaller subnets than /64 is because the adresses are partly dynamically generated. /64 is the smallest subnet where we consider the chance for collision appropriately small.

in order to support smaller subnets, we need to support setting allocation pools for the Spec.Options.Net6 parameter. which is possible to do with an appropriate size limitation (meaning that the _maximum_  of the IPv6 allocation pool needs to be restricted)
so it is a possible enhancement

ArunN85:
Yes..  i get /120 network.

topranks:
You shouldn't be using subnets of that size dude.  Sorry if this is no help, I don't think it is a good idea for the project to support things which do not comply with the standards though.

wallyatgithub:
Hi,

It is exciting to see you have done 2 merges.

So is there any more commit to occur, or, this feature is ready now?

Thanks a lot.

Levovar:
> Hi,
> 
> It is exciting to see you have done 2 merges.
> 
> So is there any more commit to occur, or, this feature is ready now?
> 
> Thanks a lot.

mostly ready, but there will be still 2 more PRs
[...]

PR #177, opened 2020-02-04
Adding IPv6 allocation pool to the APIs

Levovar:
Implements part of https://github.com/nokia/danm/issues/147

This includes adapating the client code, and working the new attributes into the webhook validation rules.
Allocation bitmask for IPv6 becomes a mandatory attribute, just as for IPv4. This means that random EUI64 generation based IPv6 allocation mode is planned to be discontinued.

A network can only store 16M of addresses, total, so Webhook makes sure this rule is never violated.
Contrary to IPv4 allocation pool we create an allocation pool only for a sub CIDR. This needs to be defined by the user, or is defaulted to CIDR containing "16M-size(IPv4 allocation pool)" addresses.

Real IPv6 allocation and freeing code is not yet changed by the PR, it only handles API changes, validation, and defaulting


Levovar:
Update: latest etcd with K8s 1.16.X only seems to be able to handle 8M addresses, rather than the previously assumed 16M
limits needed to be lowered to match this restriction, other than that PR works as intended

Issue #185, opened 2020-03-05
Svcwatcher doesn't handle concurrent updates correctly


Levovar:
**Is this a BUG REPORT or FEATURE REQUEST?**:
bug

**What happened**:
Whenever svcwatcher threads need to update the same API object concurrently (e.g. EndPoint), random race ocnditions can happen. K8s API makes sure there is always only one write going through, and the rest is declined with an OptimisticLockerrorMessage.
Whenever this happens ti is the client's job to refresh the object and retry the write operation - something which is currently not done by svcwatcher.

**What you expected to happen**:
Svcwatcher always succeeds in updating K8s API objects, even when multiple write operations are 
triggered for the same object.

**How to reproduce it**:
Create multiple svcwatcher managed services at the same time, e.g. from compex Helm charts with multiple sub-charts.

PR #187, opened 2020-03-06
Adding V6 support to svcwatcher, plus fixing concurrent updates

Levovar:
DANM svcwatcher managed EndPoints now support IPv6 addresses as well!

also correcting https://github.com/nokia/danm/issues/185
Correction is not the elegant one, but should work with the current architectural constraints.
We will infinitely re-try EndPoint update in case of conflict, but only retry for 1 sec in case of temporary network outages etc.

### DIFF
--- a/cmd/svcwatcher/svcwatcher.go
+++ b/cmd/svcwatcher/svcwatcher.go
@@ -1,124 +1,113 @@
 package main
 
 import (
-	"context"
-	"flag"
-	"k8s.io/client-go/transport"
-	"log"
-	"time"
-	"os"
-	"fmt"
-	"github.com/golang/glog"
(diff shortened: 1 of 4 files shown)


Levovar:
ok, so to record for the next generation: I ended up adding an extremely ugly concurrency handling block to all EndPoint update part
it looks to be working, but I can't stand to look at it. anyone feeling frisky enough has my blessing to refactor the whole controller :)

other than that, performance bottlenecks can happen if the same EndPoint in the same service needs to be conccurently updated, e.g. you create 100 Pod at the same time, matchin the exact same Service
we will probably handle it eventually, but it will take some seconds to update all PodIPs
IMO it should work okayish until like 100 concurrent updates, wouldn't try using it for thousand Pods / Service

PR #194, opened 2020-03-16
Make Webhook deploy CA bundle scripts executable

ckoester2:
The deployment guide calls for running a set of shell scripts
to create webhook's certificate and CA bundle. Change permissions
of these scripts so that they are actually executable. Also,
minor documentation update that "jq" is required by these
scripts.

This is obviously trivial to fix for anyone who follows the deployment
guide and runs into the error -- but at the same time, it seems there
is no reason to not fix this in the repository.

**What type of PR is this?**
fix

One less moment of "oops"->fix->try again during the deployment
process.

NONE


Levovar:
@ckoester2 needs rebasing but otherwise sounds good, and appreciated!

ckoester2:
> @ckoester2 needs rebasing but otherwise sounds good, and appreciated!

done that now :)

PR #195, opened 2020-03-16
Grant Events Create permission to svcwatcher RBAC

ckoester2:
Seeing following error in svcwatcher log,
'User "system:serviceaccount:kube-system:svcwatcher" cannot
create resource "events" in API group "" in the namespace "kube-system"'.

Adding this API to svcwatcher RBAC.

**What type of PR is this?**
bug

NONE

### DIFF
--- a/integration/manifests/svcwatcher/0svcwatcher_rbac.yaml
+++ b/integration/manifests/svcwatcher/0svcwatcher_rbac.yaml
@@ -34,6 +34,14 @@ rules:
   - watch
   - get
 - apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - create
+  - update
+  - get
+- apiGroups:
   - "danm.k8s.io"
   resources:
   - danmeps


integration/manifests/svcwatcher/0svcwatcher_rbac.yaml
```diff
@@ -33,6 +33,12 @@ rules:
   - list
   - watch
   - get
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
```
Levovar: when we fixed -in another :) - upstream we found we need to add more verbs for the leader election library to work as intended:
https://gerrit.akraino.org/r/c/ta/caas-security/+/2084/4/rbac_manifests/svcwatcher-rbac-config.yml

ckoester2: Added update and get here, too

PR #190, opened 2020-03-14
Migrate to go modules

eMGabriel:
Change build.sh to speed up build and remove unfixed dependencies
Add hack/tools.go giving dev-tools version fixing via go.mod (e.g. code-generator)
Remove glide
Cleanup gitignore

**What type of PR is this?**
cleanup

**What does this PR give to us**:

**Which issue(s) this PR fixes** *(in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #196, opened 2020-03-17
Refactor build process into single Docker build

ckoester2:
Refactor "build_danm.sh" script, Dockerfile and scm/build/build.sh
such that:

 - The build process happens entirely inside a container. There
   is no longer a requirement to have "go" installed on the host
   machine at all.

 - There is no longer a dependency on the build process to have
   any particular directory structure or volumes mounted from the
   host.

 - There is a single "one-stop" build script. There is no longer
   a requirement to build a build container, then run that build
   container, then copy binaries into another (integration/)
   directory and start another container build.

In addition, this change also packages up the CNI binaries
(danm & fakeipam) as well as the 3rd party macvlan, into a
container that then runs as part of a DaemonSet, moving the
binaries into place.

Last (and admittedly unrelated, but at very little extra cost),
this change also builds macvlan and deploys it along with the
DANM CNI. The source was already part of the build container,
so there is no extra cost there, and the additional time to
compile (and size added to the build container) are minimal.

**What type of PR is this?**
design

A cleaner build process (especially for users, not developers) that:

  - Has no dependency on golang outside of the build container. (It seems that
    even before this change, the was actually no compiling or running of go binaries
    [other than the statically compiled plugins] happening on the host, but golang
    was being used as a kind of drop-in replacement for "git clone"). BUT the previous
    Dockerfile was mounting volumes with reference to $GOPATH, so was exposing
    a technically unnecessary dependency on golang on the host.

  - Has no dependency on host network or host volume. Can be run on pretty much
    any machine that has docker (or buildah) installed and is able to execute a shell script.
    All data remains contained within the git workspace, and there is no requirement to
    export files back from the container into the host filesystem.

  - SHOULD (admittedly, untested) work with buildah again. The previous build was broken
    for buildah, because commit 68703134 change the mount directory from /go to /usr/local/go
    but did so only for the docker build. This diff merges build_docker and build_buildah into
    the same script, thus lowering the risk that one of the two build platforms gets missed
    in future changes.

  - Is considerably easier to use. Rather than running a build script, copying three resulting binaries into
    three integration directories, and running three more "docker build" commands,
    this process merges all of the above into a single build step.

  - Easier deployment of the CNI binaries. They, too, get packaged into a container image,
    and that image gets deployed as a DaemonSet, along with a shell script that copies these
    images into /opt/cni/bin. This eliminates another step of manually moving files around.

    That DaemonSet approach would also allow for the kubectl and 00-danm.conf files to be deployed
    by the DaemonSet - I'm happy to look into that in a subsequent change.

  - An updated deployment-guide to reflect these changes.

Fixes #

Will ping you in Slack channel too.

While functional, there's a few things to consider still:

  - The code, as-is, while not having changed any golang code, will likely break the unit
    tests (or require the test to pull the binaries out of the container images and place them
    into their golang directory for unit testing). I have a few thoughts on how to address
    this, but might need some discussion about how the current go developer pattern work pattern
    is, to see which approach works best.

  - Also, not sure if this breaks your CI integration in any way?

  - Finally, this will conflict with PR#190. Happy to look at how the two can be merged, after
    initial review and commenting are done :-)

```release-note
Build and deployment process changed. See updated deployment-guide.md.
```


deployment-guide.md
```diff
@@ -20,44 +20,27 @@ To be able to do that, your development environment shall already have Docker da
...
 
-The result will be 6, statically linked binaries put into your $GOPATH/bin directory.
+The result will four container images:
 
-- danm
+- `danm-cni-plugins`: This image contains the core CNI plugins (`danm`, `fakeipam`) as well as the `macvtap`
+  plugin. It will be deployed as a DaemonSet that puts these binaries in place in each Kubernetes node.
 
-- fakeipam
+- `netwatcher`: This image will be used by the `netwatcher` DaemonSet
```
Levovar: what we did in our product is that we ultimately created a "hyperdanm" container image, with all three binaries
I think it would be the easiest way to go

deployment-guide.md
```diff
@@ -20,44 +20,27 @@ To be able to do that, your development environment shall already have Docker da
...
 ./build_danm.sh
 ```
-This will first build the Alpine based builder container, mount the $GOPATH/src and the $GOPATH/bin directory into it, and invoke the necessary script to build all binaries inside the container.
-The builder container destroys itself once its purpose has been fulfilled.
 
-The result will be 6, statically linked binaries put into your $GOPATH/bin directory.
+The result will four container images:
 
-- danm
+- `danm-cni-plugins`: This image contains the core CNI plugins (`danm`, `fakeipam`) as well as the `macvtap`
```
Levovar: why macvtap, and not macvlan? do you use it for something? :)
we use macvlan, ipvlan and sriov mostly for control plane and data plane, so if we put some binaries inside IMO macvlan should bei ncluded instead / in addition to macvtap

integration/bootstrap_networks/production/flannel.conf
```diff
@@ -0,0 +1,7 @@
+{
+  "cniVersion": "0.3.1",
+  "type": "flannel",
+  "delegate": {
+    "isDefaultGateway": true
```
Levovar: I have a vague memory of Flannel not properly installing default routes if "hairpinMode": true was not defined, but maybe it is not true anymore
can include it anyway, just in-case

scm/build/build.sh
```diff
@@ -1,15 +1,26 @@
...
+
 go install -a -ldflags "-extldflags '-static' -X main.version=${LATEST_TAG} -X main.commitHash=${COMMIT_HASH}" github.com/nokia/danm/cmd/danm
 go install -a -ldflags "-extldflags '-static' -X main.version=${LATEST_TAG} -X main.commitHash=${COMMIT_HASH}" github.com/nokia/danm/cmd/netwatcher
 go install -a -ldflags "-extldflags '-static' -X main.version=${LATEST_TAG} -X main.commitHash=${COMMIT_HASH}" github.com/nokia/danm/cmd/svcwatcher
 go install -a -ldflags "-extldflags '-static' -X main.version=${LATEST_TAG} -X main.commitHash=${COMMIT_HASH}" github.com/nokia/danm/cmd/webhook
 go install -a -ldflags "-extldflags '-static'" github.com/nokia/danm/cmd/fakeipam
-go install -a -ldflags "-extldflags '-static'" github.com/nokia/danm/cmd/cnitest
\ No newline at end of file
+go install -a -ldflags "-extldflags '-static'" github.com/nokia/danm/cmd/cnitest
+go install -a -ldflags "-extldflags '-static'" github.com/containernetworking/plugins/plugins/main/macvlan
```
Levovar: ah okay, so you do meant macvlan :D
then maybe it is just a documentation mismatch, but the two are not the same IMO (e.g. https://github.com/maiqueb/macvtap-cni , and this is a "normal" bridged network interface whereas macvlan creates macvlan slaves directly attached to the host device)

Levovar:
> That DaemonSet approach would also allow for the kubectl and 00-danm.conf files to be deployed
by the DaemonSet - I'm happy to look into that in a subsequent change.

would very much appreciate it. that would effectively close our longest outstanding issue https://github.com/nokia/danm/issues/35

> The code, as-is, while not having changed any golang code, will likely break the unit
tests (or require the test to pull the binaries out of the container images and place them
into their golang directory for unit testing). I have a few thoughts on how to address
this, but might need some discussion about how the current go developer pattern work pattern
is, to see which approach works best.

have you tried it or just an assumption? I'm not sure UTs depend on precompiled binaries in the workspace. IMO go test simply compiles and executes its own test bin from source
at least I frequently get compile errors for the production code when I execute newly written TCs and mess-up some APIs

> Also, not sure if this breaks your CI integration in any way?

Yes :)
https://travis-ci.org/github/nokia/danm/builds/663336486
but only the UTs, and only because of the missing GOPATH
if we merge the modularization first that might solve it in itself

scm/build/Dockerfile
```diff
@@ -1,24 +1,139 @@
...
+# Stage: Netwatcher
+#
+FROM alpine:latest AS netwatcher
+MAINTAINER Levente Kale <levente.kale@nokia.com>
+
+COPY --from=builder /go/bin/netwatcher /usr/local/bin/netwatcher
+
+RUN apk add --no-cache --virtual .tools curl libcap iputils  \
+ && adduser -u 147 -D -H -s /sbin/nologin danm \
+ && chown root:danm /usr/local/bin/netwatcher \
```
eMGabriel: What about doing `chown` in `COPY` like `COPY --from=builder --chown root:danm ...`
Also you have to do `adduser` before `COPY` to work

scm/build/Dockerfile
```diff
@@ -1,24 +1,139 @@
...
+ && chmod 750 /usr/local/bin/netwatcher \
```
eMGabriel: I suggest doing `chmod` in `builder` container for all binaries, then will be no need to do it in every component stage

scm/build/Dockerfile
```diff
@@ -1,24 +1,139 @@
...
+ && setcap cap_sys_ptrace,cap_sys_admin,cap_net_admin=eip /usr/local/bin/netwatcher \
+ && setcap cap_net_raw=eip /usr/sbin/arping \
```
eMGabriel: It was long ago when danm was relying on arping binary, is there any other component that depends on it?
How about doing `setcap` for needed binaries in `builder` container? Would it be set for the copied file?

scm/build/Dockerfile
```diff
@@ -1,24 +1,139 @@
...
+#
+FROM alpine:latest AS danm-cni-plugins
+
+VOLUME ["/host/cni"]
+
+RUN mkdir /cni
+
+COPY --from=builder /go/bin/danm /cni/danm
+COPY --from=builder /go/bin/fakeipam /cni/fakeipam
+COPY --from=builder /go/bin/macvlan /cni/macvlan
```
eMGabriel: These can be collapsed in one line like this:
`COPY --from=builder --chown=user:group /go/bin/danm /go/bin/fakeipam /go/bin/macvlan /cni/`

scm/build/build-deps.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh -ex
+export GOOS=linux
+
+glide install --strip-vendor
+go get -d github.com/vishvananda/netlink
+go get github.com/containernetworking/plugins/pkg/ns
+go get github.com/golang/groupcache/lru
```
eMGabriel: This file is not needed after #190 
because glide removed and dependencies added to go.mod

scm/build/cni_ds/entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+echo "Copying plugins..."
+
+# Copy files into place and atomically move into final binary name
+for plugin in /cni/*
+do
+  plugin=$(basename "${plugin}")
+
+  echo "  - ${plugin}"
+  cp -fa /cni/"${plugin}" /host/cni/_"${plugin}"
+  mv -f  /host/cni/_"${plugin}" /host/cni/"${plugin}"
```
eMGabriel: Could you clarify why first `cp` to temp name, then rename with `mv`? 🤔

deployment-guide.md
ckoester2: yes, sorry, brain fart/typo. Meant `macvlan`.

scm/build/Dockerfile
ckoester2: Quite honestly I pretty much just moved what was in the previous component Dockerfiles (eg. `integration/dockerfiles/netwatcher/Dockerfile` in this example) and used it 1:1. But I do agree that this might be an opportunity to review these tasks. Will modify as you suggested...

scm/build/build-deps.sh
ckoester2: Agreed. I think that generally there'll be some overlap with #190 , and I actally wonder if it makes sense to merge #190 first and then I'll update this change based on the current situation then?

scm/build/cni_ds/entrypoint.sh
ckoester2: Yes -- the comment on line 5 (not shown in the snipplet here, but a couple of lines up) was intended to address that :-)

**Long answer**

"mv" is an atomic operation; "cp" is not.

If someone else is attempting to execute the binary while we install it (which *is* possible in Kubernetes' declarative nature), then with "cp", there is a risk that they consume a partially-deployed binary.

Someone could have deployed multiple Kubernetes resources simultaneously; for example, someone could have `kubectl create`d the CNI daemonset, netwatcher, and svcwatcher at the same time -- in this case it is entirely possible that `svcwatcher` is already sitting and waiting to run the cni plugin, *while* the CNI plugin is being installed.

The risk is even higher after a node reboot, or in case of a CNI daemonset update (replacing the old CNI with a newer version). In this case, a whole set of apps might already be trying to (re-)start while the daemonset is restarting also (and will re-create the binaries). Again, in this case there'd be a risk that someone consumes an incomplete binary during a write-in-progress.

In both cases, this may result in unpredictable/undefined errors.

Copying the file to a temporary location and then doing a `mv` (which *is* an atomic operation), will prevent this. With `mv`, the consumer will either see no binary at all (in which case K8s will retry and wait for the plugin to be available), or a complete and sane binary - but never anything in between.


**Short answer**:

Other CNI plugins (eg. Multus <https://github.com/intel/multus-cni/blob/master/images/entrypoint.sh#L155>) do that, too ;-)

scm/build/build-deps.sh
eMGabriel: If I were in your shoes, I would do the following:

1. master-master sync between your fork and nokia/danm
2. rebase on your synced fork `master` and step by step resolve every commit conflict
3. cleanup, remove not needed content
4. `git commit --amend` this removes not needed content/files from PR
5. then `git push --force` because of rewriting commit in step 4. by `--amend` ()

For a healthy git history ✌️ 
But maybe you already know what to do. Trust your instincts!😄

scm/build/cni_ds/entrypoint.sh
eMGabriel: My guess was the same, but I was not confident enough. I am okay with it.

scm/build/build-deps.sh
ckoester2: Yes, thanks. I guess the one thing I hadn't seen, was that #190 did get merged one hour before my comment, and I somehow assumed it was still under review. With it being merged now, I'll go ahead and update this - yes, I've worked with the mechanics of git, thanks :)

scm/build/Dockerfile
ckoester2: Will remove the arping.

Unfortunately, no, xattrs (including what's set by setcap) are not preserved in a multistage build, it seems

Quick test:
```
% cat Dockerfile
FROM alpine:latest as stage1
RUN apk add --no-cache libcap attr \
 && touch test && echo "S1 BEFORE SET:" && getcap -v test && attr -l test \
 && setcap cap_net_admin=eip test  && echo "S1 AFTER SET:" && getcap -v test && attr -l test

FROM alpine:latest as stage2
COPY --from=stage1 test test
RUN apk add --no-cache libcap attr \
 && echo "STAGE2:" && getcap -v test && attr -l test
```

output (abbreviated):

```
S1 BEFORE SET:
test
S1 AFTER SET:
test = cap_net_admin+eip
Attribute "capability" has a 20 byte value for test

STAGE2:
test
```

https://github.com/moby/moby/issues/35699

Would have been nice, though...

eMGabriel: Oh dear! That means my spider-sense was warning me in the past. I did not know I was coming accross this issue while testing of PR #190 Migrate to go modules on a live system. It a pity. Thanks for investigating it. 👍

integration/bootstrap_networks/production/flannel.conf
ckoester2: Done.

deployment-guide.md
ckoester2: Could do, but how strong do you feel about that? It would reduce the size of the Dockerfile. it would marginally reduce the total consumed space for container images, but I have to admit I like the idea of having three separate images for the separate pods. Keeps things tidy somehow; each pod contains only the binary that it actually needs.

For the user, it should not make a difference anyway -- their `kubectl apply` look all the same :)

integration/bootstrap_networks/production/flannel.conf
ckoester2: Done.

scm/build/Dockerfile
ckoester2: Ok, cleaned up the Docker stage builds a bit... probably best to re-review the entire file as a whole :)

In short,

- moved all the chown/chmod into the build stage;
- created an intermitted stage which is just the Alpine image and adds the user (and then acts as a parent stage for netwatcher, svcwatcher, webhook)
- removed arping,
- removed a whole bunch of apk-add/apk-del that seemed to no longer serve any purpose

scm/build/Dockerfile
ckoester2: Done.

scm/build/Dockerfile
ckoester2: Done.

scm/build/build-deps.sh
ckoester2: Done.. and collapsed build-deps/build into a single stage again

ckoester2:
With the go upgrade and change to go modules, no longer including `macvlan` in the image here after all. Now, it would take extra downloads and extra effort, so decided to keep that separate.

deployment-guide.md
eMGabriel: How about supporting both and avoid tying the users' hands?

build_danm.sh
```diff
@@ -30,12 +53,78 @@ error_handler()
...
+# by one target in the multi-stage Dockerfile.
+#
+for plugin in ${build_targets[@]}
+do
+  echo Building: ${plugin}, version ${COMMIT_HASH}
+  ${BUILD_COMMAND} \
+    ${EXTRA_BUILD_ARGS} \
+    ${FIRST_BUILD_EXTRA_BUILD_ARGS} \
+    --build-arg LATEST_TAG=${LATEST_TAG} \
+    --build-arg COMMIT_HASH=${COMMIT_HASH} \
```
eMGabriel: `--build-arg` can be specified with variable name only, like `--build-arg LATEST_TAG` and its value will be loaded from the environment. I think it would be cleaner this way

ckoester2: Could do, but keep in mind that that would require `export`ing these variables from the build script. It certainly would be another option -- not sure if it'd be all that much better though.

deployment-guide.md
ckoester2: Are you suggesting to build both, and let the user chose which to install?

Or to offer two versions of the build script (or one script with conditionals), and two paths of documentation?

eMGabriel: I suggest building both the `single binary per image` and the `multi binary per image` method, because as @Levovar mentioned in our product we prefer the `multi binary image` method, butI see other users (including You also?🌝) would prefer single binary images.
> Is this something that could potentially be done in a future enhancement?

Yes, of course. If you are not interested in the multi binary image method, just leave it to me, I will add it later after your PR is merged.

build_danm.sh
eMGabriel: Okay, then leave it as is.

deployment-guide.md
ckoester2: I'm happy to look at it over the next few days - but if you guys can live with the current version, I'd probably prefer to merge this as-is, and then work on it as an incremental change that adds the option to build a second form factor.

Levovar: yeah just an idea, as you can see I feel so strongly about it I also haven't brought it over to upstream in the last 6 months :)

PR #199, opened 2020-03-21
Fix possible nil derefernce in metacni error handling

### LINKED ISSUE #198: CNI plugin error handling seems broken

bug

**What happened**:

It looks like since recent commits (fd4dd7 or around that time), errors during CNI plugin execution result in a CNI crash which prevents the CNI from reporting a useful error.

I don't have a complete list of possible error scenarios (and no idea if there is no such list), but a few scenarios come to mind that I'm reasonably sure were gracefully handled with a meaningful error message before, and now just result in an unspecific error. My overall impression is that any downstream CNI plugin error, results in the DANM CNI crashing rather than cascading the error upstream.

Downstream CNI plugin errors to be handled gracefully and reported upstream, so that logs / `kubectl describe` output to have useful information.

**How to reproduce it**:

Consider the following configuration (somewhat abbreviated):

```
root@debian:/home/debian/sandbox/dup-ip# cat *
---
apiVersion: danm.k8s.io/v1
kind: ClusterNetwork
metadata:
  name: testnet
spec:
  NetworkID: testnet
  NetworkType: macvlan
  Options:
    host_device: dummy1
    container_prefix: eth
    cidr: 10.250.1.0/24
    allocation_pool: {start: 10.250.1.100, end: 10.250.1.200}
---
apiVersion: v1
kind: Pod
metadata:
  annotations:
    danm.k8s.io/interfaces: '[{"clusterNetwork":"testnet", "ip":"10.250.1.100"}]'
spec:
  containers:
  - name: sleep
    image: sleep-infinity
---
apiVersion: v1
kind: Pod
metadata:
  annotations:
    danm.k8s.io/interfaces: '[{"clusterNetwork":"testnet", "ip":"10.250.1.100"}]'
  name: pod2
spec:
  containers:
  - name: sleep
    image: sleep-infinity
```

This is clearly an invalid configuration - both pods share the same network and same IP address.

Running the above, results in:

```
$ kubectl get pods
NAME   READY   STATUS              RESTARTS   AGE
pod1   0/1     ContainerCreating   0          6m30s
pod2   1/1     Running             0          6m30s
```

which is somewhat expected. But then, trying to troubleshoot this:

```
$ kubectl describe pod pod1
...
Events:
  Type     Reason                  Age                   From               Message
  ----     ------                  ----                  ----               -------
  Normal   Scheduled               <unknown>             default-scheduler  Successfully assigned default/pod1 to debian
  Warning  FailedCreatePodSandBox  6m46s                 kubelet, debian    Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "afff4e92d21d889ea1afa11f36568e95a43cd63cc290a416e1ece90ef241b46a": netplugin failed with no error message
```

The "netplugin failed with no error message" message is where this bug report comes in. I am sure I ran into this error before by mistake, and *did* see a meaningful error pointing at the duplicate IP address.

**Environment**:
- DANM version (use `danm -version`): v4.1.0-42-gfd4dd71
- Kubernetes version (use `kubectl version`): 1.17.3
- DANM configuration (K8s manifests, kubeconfig files, CNI config file):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Others:

ckoester2:
Fix issue where in meta CNI where if an error during createNic()
might result in a possible nil pointer dereference during error
handling

**What type of PR is this?**
bug

A fix for issue #198 . Maybe not in the right way - please review and let me know.

**Which issue(s) this PR fixes**
Fixes #198 

NONE

### DIFF
--- a/pkg/metacni/metacni.go
+++ b/pkg/metacni/metacni.go
@@ -343,7 +343,7 @@ func createNic(syncher *syncher.Syncher, danmClient danmclientset.Interface, ifa
     if ep != nil {
       danmep.DeleteDanmEp(danmClient, ep, netInfo)
     }
-    syncher.PushResult(ep.Spec.NetworkName, err, nil)
+    syncher.PushResult(netInfo.ObjectMeta.Name, err, nil)
     return
   }
   var cniResult *current.Result


ckoester2:
@Levovar  - I see that you specifically changed this **from** `netInfo.ObjectMeta.Name` **to** `ep.Spec.NetworkName` in commit 27141f92678b4b175704ce1a90b91413f4ec6902. So not sure if my changing this back is the right way, or if this is going to re-introduce an earlier issue that you were trying to fix.

But with a bit of poking around, it certainly seems to me that `ep` is Nil in at least some error cases (for example, in the configuration described in issue #198), and so then, derefencing `ep->Spec` creates a panic.

Levovar:
yep EP can be definitely nil there, that's why I even added the if before DeleteDanmep, just failed to notice the sycnher call below can also core
I think I only wanted the code to look uniform. I'm not sure if network cna be nil at that point, digged a little bit more and I don't think so

Issue #191, opened 2020-03-16
Svcwatcher shouldn't try and deal with Endpoints, Pods, Services, and DanmEps which are not in the same namespace as the changed object


Levovar:
However currently this is unoptimized, and it does.
This is dues to using generic Lister K8s client go API: https://github.com/kubernetes/client-go/blob/kubernetes-1.11.10/listers/core/v1/service.go#L31
rather than the NamespaceLister one: https://github.com/kubernetes/client-go/blob/kubernetes-1.11.10/listers/core/v1/service.go#L33

it does not result in functional error, but can generate misleading error messages, and unnecessary event processing

PR #200, opened 2020-03-23
Fixing too generic Endpoint, DanmEp, Services, and Pod interrogation in svcwatcher

Levovar:
Fixing https://github.com/nokia/danm/issues/191
When handling an API event related to thes objects, svcwatcher unnecessrily
interrogated ALL existing objects.
It is fixed now by always using the namespace awarre lister client API instead the generic.


Levovar:
Optimization works, and results in an actually noticeable increase in event processing speed

PR #206, opened 2020-04-03
Fix CRD IPv4/IPv6 CIDR validation pattern to strict

eMGabriel:
Kubernetes provides pre-defined [string validation formats](https://github.com/kubernetes/apiextensions-apiserver/blob/release-1.17/pkg/apis/apiextensions/v1/types_jsonschema.go#L27)
format: cidr - validates that string is valid IPv4 or IPv6 CIDR
Because we want to allow only one of them for each field value, adding a filter pattern solves this
In IPv4 CIDR case:
pattern: '^\d+\.' - pattern matches first IPv4 block + dot, first IPv6 block cannot be followed by dot
In IPv6 CIDR case:
pattern: ':' - pattern matches at least one colon in IPv6, there is no colon in IPv4

**What type of PR is this?**
bug

Fixes an error during POD creation because CRD schema validation is erroneous for IPv6 related string fields in DanmNet, ClusternNetwork, TenantNetwork.
Causes following error:
`
<...> : Invalid value: "": "spec.Options.allocation_pool_v6.cidr" must validate one and only one schema (oneOf). Found none valid
<...> spec.Options.allocation_pool_v6.cidr: Invalid value: "": spec.Options.allocation_pool_v6.cidr in body should match '([0-9a-fA-F]*:)+[\d.]*/\d+'
`

None

```release-note
The schema validation field 'format:' is officially supported from kubernetes 1.16, older releases may not support it.
```

### DIFF
--- a/integration/crds/lightweight/DanmNet.yaml
+++ b/integration/crds/lightweight/DanmNet.yaml
@@ -32,7 +32,7 @@ spec:
                     maxLength: 0
                   - type: string
                     format: cidr
-                    pattern: '[\d./]+'
+                    pattern: '^\d+\.'
                 allocation_pool:
                   properties:
                     start:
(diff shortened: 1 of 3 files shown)


Levovar:
Regex is hard :D

PR #207, opened 2020-04-08
Add custom ListAndWatch error handler to netwatcher Informer

Levovar:
Based on recent client enhancement: https://github.com/kubernetes/kubernetes/pull/87329

When an existing watcher thread fails with an unexpected error, netwatcher will shut itself down using this custom handler.
Kubernetes will promptly restart it, re-initializing the watches thus ensuring overall netwatcher HA - even if it comes at the expensive of some Pod restarting..
It is observed that the client library provided Informer HA cannot tolerate prolonged API server failures, so this enhancement makes sure netwatcher always comes back when the API server returns.


pkg/netcontrol/netcontrol.go
```diff
@@ -77,6 +85,19 @@ func (netWatcher *NetWatcher) Run(stopCh *chan struct{}) {
...
+func (netWatcher *NetWatcher) WatchErrorHandler(r *cache.Reflector, err error) {
+	if apierrors.IsResourceExpired(err) || apierrors.IsGone(err) || err == io.EOF {
+    log.Println("INFO: One of the API watchers closed gracefully, re-establishing connection")
+    return
+  }
+  //The default K8s client retry mechanism expires after a certain amount of time, and just gives-up
+  //It is better to shutdown the whole process now and freshly re-build the watchers, than risking becoming a permanent zombie
+  close(*netWatcher.StopChan)
+  //Give some time for gracefully terminating the connections
+  time.Sleep(5*time.Seconds)
```
eMGabriel, line 97: it is time.Second

Levovar: I know, but thanks
it just anyway doesn't matter until go mod stuff is taken care of :)

eMGabriel: In client-go this commit holds the needed functionality: [ccd5becdffb7fd8006e31341baaaacd14db2dcb7](https://github.com/kubernetes/client-go/commit/ccd5becdffb7fd8006e31341baaaacd14db2dcb7)
Try fixing client-go to this commit in go.mod by issuing this command in module root:
`go get k8s.io/client-go@ccd5becdffb7fd8006e31341baaaacd14db2dcb7`
That only updates client-go without its dependencies. If that is not enough and client-go fails to build because of mismatching dependencies, then try to update its dependencies with
`go get -u k8s.io/client-go@ccd5becdffb7fd8006e31341baaaacd14db2dcb7`

Also there is possibility to bind client-go dependency to current head with `go get k8s.io/client-go@master`

Levovar: i literally tried updating ALL dependencies on Thursday, nothing seemed to work
feel free to experiment yourself as well, from my side I will just rather wait for the next tag

Levovar: thanks for the PR :)

Levovar:
This might not be what the creators behind the API intended, but it seems to work beautifully

PR #212, opened 2020-04-26
Add installer job

### LINKED ISSUE #35: Danm installer and popularization

carstenkoester:
Adds an alternate method of installing by running a Kubernetes Job.
See file "deployment-installer-job.md" in this commit for a detailed
description of the job.

**What type of PR is this?**
feature

A simple installer job

Fixes #35 

With this change in place, it might actually make sense to pre-build DANM and upload to DockerHub? That would even further simplify the deployment.

A Helm chart might be yet another option, but I don't know of a good way to do the Kubernetes CA integration (issuing the CSR, getting it signed) by Helm alone. It seems that that would still require a job of sort; some other users seem to be using https://github.com/jet/kube-webhook-certgen for that purpose?). In any case, this provides one option that works in environments where Helm is not available, too.

In hindsight, I feel rather silly for having mounted the installer configmap as file mounts (rather than mounting the configmap as environment variables). At the start, I thought there'd be more cases of directly manipulating files -- but they all disappeared over the time, and what's left now, is largely a construct where most mounted files are directly read into environment variables.

Note that this change currently leaves two versions of netwatcher/webhook/cni-daemonset/svcwatcher manifests -- the original example files, and a `.tmpl` file for automated parsing. There may be an opportunity to remove the "manual deployment" example files at a later date.

```release-note
See deployment-installer-job.md for an alternative (simpler, but currently experimental) method
to deploy.
```

### DIFF
--- a/deployment-installer-job.md
+++ b/deployment-installer-job.md
@@ -1,0 +1,84 @@
+# Deployment via installer job
+
+
+## TL;DR;
+
+```
+${EDITOR} integration/install/danm-installer-config.yaml
+kubectl apply -f integration/install
+```
+
+## Installer job
+
+This (currently experimental) method installs DANM using a Kubernetes Job.
+
+Just like the manual deployment, it assumes that a previous CNI (also referred to
+as "bootstrap CNI") is already installed. In the setup deployed by this installer,
+the bootstrap CNI will both be used by DANM components themselves (ie. netwatcher
+and svcwatcher will utilize that bootstrap CNI for their own network connectivity),
+as well as being configured as a DanmNet or ClusterNetwork with the name "default",
+that will be used by any Kubernetes applications without a `danm.k8s.io` annotation.
+
+Please be aware that the existing (bootstrap) CNI configuration must be a single
+CNI, *not* a list of CNIs. This means that in your CNI configuration directory,
+`/etc/cni/net.d`, there should be an existing file with a `.conf` extension, often
+named something like `10-flannel.conf` or `10-calico.conf`.
+
+If there is a file with a `.conflist` extension (such as `10-calico.conflist`), then
+that is a chained list of multiple CNIs. DANM does not currently support using
+such a `.conflist` chain as a bootstrap network. Depending on your setup, you may be
+able to to extract only the first CNI from the list using a command such as the
+following:
+
+```
+jq -M '{ name: .name, cniVersion: .cniVersioni} + .plugins[0]' \
+  /etc/cni/net.d/${EXISTING_CONFLIST_FILE}.conflist \
+  > /etc/cni/net.d/${FIRST_PLUGIN_FROM_LIST_CONFIG_FILE}.conf
+```
+
+Either way, please be sure that you have a functional `/etc/cni/net.d/*.conf` CNI
+configuration before proceeding, and know the name of that `.conf` file.
+
+
+### Configuration file (configMap)
+
+This file will need modification to match your setup.
+
+Please review/edit `integration/install/danm-installer-config.yaml`.
+
+
+### Installer Job resource
+
+This file will need modification only if the installation container needs to be
+pulled from an external registry. If this is the case, then please review/edit
+`integration/install/danm-installer.yaml`.
+
+If you have built DANM locally and do not need to pull images, this file does not
+need updating.
+
+
+### Deploying the installer
+
+```
+kubectl apply -f integration/install
+```
+
+
+### Watching installer progress
+
+After applying the installer CRD, in `kubectl get pods -n kube-system` you should
+first see a `danm-installer-*` pod starting, and shortly after, the
+`danm-cni` and `netwatcher` daemonsets, `svcwatcher`, and `danm-webhook-deployment`
+pods.
+
+The `danm-installer-*` pod should end up in "Completed" status - if not, please check
+the pod logs for any errors.
+
+
+### Cleaning up (optional)
+
+After the installer pod ran to completion, you can remove the installer itself:
+
+```
+kubectl delete -f integration/install
+```
--- a/integration/install/danm-installer-config.yaml
+++ b/integration/install/danm-installer-config.yaml
@@ -1,0 +1,109 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: kube-system
+  name: danm-installer-config
+data:
+  #
+  # DANM deployment mode. This MUST be either "lightweight" or "production". Please see the DANM user guide for
+  # details on the two modes.
+  #
+  deploy_mode: production
+
+  #
+  # CNI configuration directory. Typically, this is "/etc/cni/net.d". This is the directory where your
+  # current (bootstrap) CNI configuration is located, too.
+  #
+  cni_dir: /etc/cni/net.d
+
+  #
+  # CNI naming scheme. See section "Naming container interfaces" in the user guide for a more detailed
+  # discussion. Set this parameter to "legacy" if you wish container interface names to be set exactly
+  # according to DanmNet.Spec.Options.container_prefix, or to an empty string if you wish the first
+  # interface to always be named "eth0".
+  #
+  cni_naming_scheme: ""
+
+  #
+  # [OPTIONAL] Kubernetes API Root CA certificate. If left blank, the installer will obtain the
+  # API server certificate from the Kubernetes API. Note, however, that placing a certificate here
+  # is technically more secure (as it provides external verification of the CA certificate, rather
+  # than blindly trusting the certificate that we see from the server) and also more future-proof
+  # if the individual API server's certificate ever were to change in the future, to put the Root
+  # CA certificate here. You can obtain this, for example, by running:
+  #
+  # kubectl config view \
+  #   --flatten \
+  #   -o jsonpath='{.clusters[0].cluster.certificate-authority-data} \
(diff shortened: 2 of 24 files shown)


Levovar:
yeah there is definitely a need to auto-create and store images on gitHub, but as far as I understood in the past there are some legal roadblocks to take care of first (regardless how silly it sounds)
summoning @CsatariGergely : any updates on storing binaries / images on Dockerhub? it is literally the only remaining roadblock of the auto installer now

deployment-installer-job.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Installer job
+
+This (currently experimental) method installs DANM using a Kubernetes Job.
+
+Just like the manual deployment, it assumes that a previous CNI (also referred to
+as "bootstrap CNI") is already installed. In the setup deployed by this installer,
+the bootstrap CNI will both be used by DANM components themselves (ie. netwatcher
+and svcwatcher will utilize that bootstrap CNI for their own network connectivity),
+as well as being configured as a DanmNet or Cluster-network with the name "default",
```
Levovar, line 19: ClusterNetwork

deployment-installer-job.md
```diff
@@ -0,0 +1,84 @@
...
+that will be used by any Kubernetes applications without a `danm.k8s.io` annotation.
+
+Please be aware that the existing (bootstrap) CNI configuration must be a single
+CNI, *not* a list of CNIs. This means that in your CNI configuration directory,
+`/etc/cni/net.d`, there should be an existing file with a `.conf` extension, often
+named something like `10-flannel.conf` or `10-calico.conf`.
+
+If there is a file with a `.conflist` extension (such as `10-calico.conflist`), then
+that is a chained list of multiple CNIs. DANM does not currently support using using
```
Levovar, line 28: using using

deployment-installer-job.md
```diff
@@ -0,0 +1,84 @@
...
+
+```
+kubectl apply -f integration/install
+```
+
+
+### Watching installer progress
+
+After applying the installer CRD, in `kubectl get pods -n kube-system` you should
+first see see a `danm-installer-*` pod starting, and shortly after, the
```
Levovar, line 70: see see

deployment-installer-job.md
```diff
@@ -0,0 +1,84 @@
...
+`danm-cni` and `netwatcher` daemonsets, `svcwatcher`, and `danm-webhook-deployment`
```
Levovar, line 71: Does DANM-CNI need to be a daemonset? Shouldn't it be also a Job?
I mean what does the CNI Pod do once it installed the CNI binaries on the host?

integration/install/danm-installer-config.yaml
```diff
@@ -0,0 +1,89 @@
...
+  #
+  # CNI configuration directory. Typically, this is "/etc/cni/net.d". This is the directory where your
+  # current (bootstrap) CNI configuration is located, too.
+  #
+  cni_dir: /etc/cni/net.d
+
+  #
+  # CNI naming scheme. See section "Naming container interfaces" in the user guide for a more detailed
+  # discussion. Set this parameter to "legacy" if you wish container interface names to be set exactly
+  # according to DanmNet.Spec.Options.container_prefix, or to an empty string if you with the first
```
Levovar, line 23: with -> wish

integration/install/danm-installer-config.yaml
```diff
@@ -0,0 +1,89 @@
...
+#    ...
+#    VGhpcyBpcyB0aGUgZW5kIG9mIGEgcGxhY2Vob2xkZXIgc3RyaW5nCg==
+#    -----END CERTIFICATE-----
+
+  #
+  # This is the type of the CNI plugin used for your default (bootstrap) network.
+  # The value can also be found in the "type" field of your bootstrap CNI configuration
+  # file, eg. "cat ${cni_dir}/${default_cni_network_id}.conf | jq -Mr '.type'"
+  #
+  default_cni_type: calico
```
Levovar, line 53: does this need to be a manually provided parameter? I think it could be grepped from the default CNI config, decreasing the chance of manual misconfiguration happening

deployment-installer-job.md
carstenkoester: The daemonset does absolutely nothing (it runs `sleep infinity`) after it completes.

However, a job runs once per cluster, and what we really need, is to run once *per node*. I guess what we'd ideally need, is some sort of construct that, just like a job, "runs to completion" but does so once per node. I don't think such construct exists, unfortunately.

There's also the scenarios of node upgrades, or nodes being added to the cluster. We need to make sure that if a new node joins the cluster, the CNI plugin is installed on that node -- even if that happens after the installer job is done.

It seems to me that a daemonset (even with the ugly hack of sleeping infinitely) is the best way to achieve this -- this also seems to be the approach that other CNIs have chosen.

integration/install/danm-installer-config.yaml
carstenkoester: Depends on the default CNI. Calico has a kubeconfig that we could fetch this from; Flannel, on the other hand, does not.

We can (and do, in my proposed code) fetch the API server certificate from the API if no certificate is provided. So we can do without it -- but technically, the API server certificate is not the same as the Root CA certificate. I'm not aware of any good way to fetch the Root CA certificate programmatically, "from the inside".

carstenkoester:
Thanks for the review - and sorry for the many typos. Will fix them....

deployment-installer-job.md
Levovar: okay, fair points :)

integration/install/danm-installer-config.yaml
Levovar: I didn't mean the CA, I only meant the name of the CNI binary which needs to be invoked for this network. I guess that is the only thing this parameter would contorl right?
and that is a mandatory attribute in a CNI config file, so I thought we could just grep it

carstenkoester: Err yes, sorry, I should have learned to read. You're right, we can probably fetch this from the CNI config file... I'll look into that.

carstenkoester: So this isn't completely trivial, but certainly doable.

The reason it is not completely trivial, is because the installer job is given the CNI configuration file directory through a configmap (`/etc/cni/net.d`), which it then uses to construct the volume mounts for the DaemonSet -- but the installer job itself, does not have `/etc/cni/net.d` mounted. For the installer to mount this, it would mean that the installer's own podspec also needs to have that volume attached. That is easy to do, of course, but would require that the user has to provide the CNI configuration path twice -- once inside the configMap, and once inside the installer podspec. This means we're trading one bit of extra work (providing the cni type) against another one (duplicating the mount directory).
I guess this is one of the places where Helm would help.

Another option is that the installer job could query this from the daemonsets. That is, the daemonsets (which have the directory mounted) could expose that information, either by a log message or by an API of sorts -- where the API could be, in the most basic, most crude scenario, the installer doing a `kubectl exec cat /host/cni/net.d/${filename}` on one of the daemonset pods. This should be fairly easy to implement, but requires some sequencing -- the installer needs to wait for the DS to be alive, *then* fetch the CNI type from one member, and *then* render the default network config. Still, this would probably be the simplest solution that does not require any additional input or duplication from the user.


As a side note, I also pushed another change right now that allows the DANM installer to *write* the boostrap CNI configuration file, rather than assuming it pre-exists. (I found that in pretty much all my Kubernetes deployments, the default CNI configuration that was installed when the cluster is created, is a .conflist - so I find myself having to deploy a different config time after time.) That, of course, solves the problem in its own way: If the installer knows the bootstrap CNI config because it is creating it, then we could simply glean the type from the config we were given. But IMHO this feature (DANM installer writing the bootstrap CNI config) should remain optional, not required by design....

Levovar: yep, you are right. totally does not worth the effort :)

Levovar:
change looks good to me, but brakes the CI:
https://travis-ci.org/github/nokia/danm/builds/682541042

carstenkoester:
> change looks good to me, but brakes the CI:
> https://travis-ci.org/github/nokia/danm/builds/682541042

Hmmm... turns out that if a shell script ends on "[ condition ] && /do/when/condition/was/true", then if the condition is *not* met, the non-zero RC of the condition becomes the shell script's exit code. As was the case with the `[ -n "${IMAGE_PUSH}" ] && ${PUSH_COMMAND} ...` syntax.

Pushed a new change that should fix it...

Issue #175, opened 2020-02-04
Enhance Fortville + DPDK support by adding extra information on the dummy kernel interface


Levovar:
When an application uses a device not managed by the kernel, DANM creates a dummy kernel interface in its place to communicate CaaS allocations to the application.
This was implemented in: https://github.com/nokia/danm/pull/155
However currently we only convey IP address, and IP routes.

The support should be expanded to cover:
- VLAN tag
- SR-IOV PCI address
as well.

Both are possible with a little trickery. the end result could look something like this:
[root@crf-dev ~]# ip -d link show dpdk1
64861: dpdk1@dumdum: <BROADCAST,NOARP,M-DOWN> mtu 1500 qdisc noop state DOWN mode DEFAULT qlen 1000
    link/ether fa:d3:bf:c3:f9:68 brd ff:ff:ff:ff:ff:ff promiscuity 0
    **vlan** protocol 802.1Q id **100** <REORDER_HDR> addrgenmode eui64
    alias **0000:17:03.5**

krsna1729:
Regarding SR-IOV PCI address - 

Is this to disambiguate which DPDK interface is which network interface? Related to discussion here https://github.com/intel/sriov-cni/issues/37 by setting the MAC address of the virtual device same as the VF, we were able to figure out the PCI/DPDK port id related to the network interface. [bess example](https://github.com/omec-project/upf-epc/blob/5be38e28052778994a2580890fe43df7acbf7b24/conf/spgwu.bess#L136-L144)

Levovar:
@krsna1729 that sounds like a great idea, thank you for the pointer and the examples!

krsna1729:
Do both, i.e, add the alias too. Its handy so that we dont have to scan and attach to all dpdk ports, instead explicitly attach the right PCI.

Setting the MAC will help for use-case i mentioned in the other issue where kernel will craft responses and DPDK app sends it out.

Levovar:
@krsna1729 with #214 merged dummy interfaces now look like this, conveying VLAN, PCI ID and even MAC address of the VFIO bound VF.
Let me know if you would like to see anything extra in addition, but otherwise I'll close this ticket :)
# nsenter -t 10885 -n ip -d a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
[...]

PR #214, opened 2020-04-27
Enhancing dummy interface capabilities with VLAN and PCI ID information

Levovar:
solves https://github.com/nokia/danm/issues/175
When the VFIO bound VF is tagged with a VLAN, we create an extra VLAN IF on top of the dummy with the same tag.
The VF's PCI ID is added to the dummy (VLAN too) as an Alias.
This mapping should be enough for applications to map allocations to the proper device.


Levovar:
added possible way of creatng dummy interface with the VFs original MAC address

also fixed a regression related to SR-IOV handling... well... being totally broken for a time now
oopsie

Issue #228, opened 2020-07-22
Custom route rule is added to wrong interface (in case of multiple DanmNets with same CIDR)


TothFerenc:
**Is this a BUG REPORT or FEATURE REQUEST?**:
bug

**What happened**:
In case there are multiple DanmNets which owns the same CIDR (only the start-end ranges are different), custom route definitions may be added to wrong interface. The same happens if the custom route rule is defined by either in DanmNet definition as `route` or in Pod annotation as `proute`. 

**What you expected to happen**:
Proper interface to be selected for such custom route rules.

**How to reproduce it**:
Sample DanmNet definitions:
```
---
apiVersion: danm.k8s.io/v1
kind: DanmNet
metadata:
  name: sriov-a
  namespace: kube-system
spec:
  NetworkID: sriov-a
  NetworkType: sriov
  Options:
    device_pool: nokia.k8s.io/sriov_ens1f0
    cidr: 10.10.10.0/24
    allocation_pool:
      start: 10.10.10.10
      end: 10.10.10.19
    rt_tables: 101
    vlan: 1301
---
apiVersion: danm.k8s.io/v1
kind: DanmNet
metadata:
  name: sriov-b
  namespace: kube-system
spec:
  NetworkID: sriov-b
  NetworkType: sriov
  Options:
    device_pool: nokia.k8s.io/sriov_ens1f1
    cidr: 10.10.10.0/24
    allocation_pool:
      start: 10.10.10.20
      end: 10.10.10.29
    rt_tables: 102
    vlan: 1301
```
Sample DANM annotation in Pod manifest:
```
        danm.k8s.io/interfaces: |
          [
            {"network":"default", "ip":"dynamic"},
            {"network":"sriov-a", "ip":"dynamic", "proutes":{"11.11.11.0/24": "10.10.10.1", "12.12.12.0/24": "10.10.10.1"}},
            {"network":"sriov-b", "ip":"dynamic", "proutes":{"12.12.12.0/24": "10.10.10.1", "22.22.22.0/24": "10.10.10.1"}}
          ]
```
Result:
```
/ # ip route
default via 169.254.1.1 dev eth0
10.10.10.0/24 dev eth2 scope link  src 10.10.10.20
10.10.10.0/24 dev eth1 scope link  src 10.10.10.10
169.254.1.1 dev eth0 scope link
/ # ip route show table 101
11.11.11.0/24 via 10.10.10.1 dev eth2
[...]

PR #233, opened 2020-08-28
Set LinkID into the netlink Route object when creating an IP route

Levovar:
fixes https://github.com/nokia/danm/issues/228

DANM now sets the link Id into the netlink Route object, making sure the kernel attaches the IP route to the correct device.

### DIFF
--- a/pkg/danmep/ep.go
+++ b/pkg/danmep/ep.go
@@ -142,26 +142,30 @@ func addIpToLink(ip string, iface netlink.Link) error {
 
 func addIpRoutes(ep *danmtypes.DanmEp, dnet *danmtypes.DanmNet) error {
   defaultRoutingTable := 0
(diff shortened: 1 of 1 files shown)


TothFerenc:
Tested with the same example from #228 :
```
/ # ip r
default via 169.254.1.1 dev eth0
10.10.10.0/24 dev eth1 scope link  src 10.10.10.10
10.10.10.0/24 dev eth2 scope link  src 10.10.10.20
169.254.1.1 dev eth0 scope link
/ # ip r show table 101
11.11.11.0/24 via 10.10.10.1 dev eth1
12.12.12.0/24 via 10.10.10.1 dev eth1
/ # ip r show table 102
21.21.21.0/24 via 10.10.10.1 dev eth2
22.22.22.0/24 via 10.10.10.1 dev eth2
```
Entries in route table 101 assigned to eth1 properly.

Issue #144, opened 2019-09-12
Webhook performance issues


Levovar:
A recent week-long performance benchmarking session identified webhook component as a bottleneck.

Issues are particularly popping because of the configured UPDATE network validation hook. For some reason webhook
1: cannot handle a big number of concurrent calls, resulting in service outage
2: cannot truly recover from service outage
3: takes very long to handle a big number of concurrent requests (longer than it takes K8s API server to persist changes into etcd)

Until the internal bottlenecks are identified the UPDATE netvalidation hook shall be disabled.
Webhook works okay under low stress, so CREATE and DELETE workflows are still safe to be used.

Levovar:
Notes related to the issue:
- https://github.com/nokia/danm/blob/master/pkg/confman/confman.go#L33 this cores in case the tconf does not contain an Alloc field, e.e.g because it was created before the webhook
- possible solution could be separating out Spec.Option.Alloc into its own CRD, so updates to Alloc would not affect the network object itself

PR #245, opened 2021-01-24
Adjust webhook timeout & Fix CNI timeout error message

TothFerenc:
**What type of PR is this?**
bug

During CNI ADD/DEL operations DANM waits 30 seconds to finish all
the delegated CNI tasks and DanmEp processing. In case the webhook
is too slow to respond, DANM timeouts without knowing the reason.
The webhook REST call timeout is lowered to 25 seconds (by default
it is 30 sec), so its timeout event can arrive back to DANM, which
can log that error message.

Related to #144, which is still a valid issue.

No.


Levovar:
RIP lónyál :(

PR #255, opened 2021-05-19
Fixing CNI result Interface indexing

Levovar:
When DANM merges together multiple CNI results it needs to change the indexing of the IPs to reference the proper Interface records from the merged result, rather than using the ID from their original one
This can cause issues when DANM is invoked directly by CRIs such as containerd instead of Kubelet. CRIs do parse the returned result, and the improper indexing can result in the wrong interface getting identified by CRI as belonging to the "cluster network"

### DIFF
--- a/pkg/metacni/metacni.go
+++ b/pkg/metacni/metacni.go
@@ -225,7 +225,7 @@ func setupNetworking(args *datastructs.CniArgs) (*current.Result, error) {
     defParam := datastructs.Interface{SequenceId: 0, Ip: "dynamic",}
     err = createIface(args, danmClient, args.DefaultNetwork, defParam, syncher, allocatedDevices)
     if err != nil {
-      syncher.PushResult(args.DefaultNetwork.ObjectMeta.Name, err, nil)
+      syncher.PushResult(args.DefaultNetwork.ObjectMeta.Name, err, nil, "")
     }
   }
   for nicID, nicParams := range args.Interfaces {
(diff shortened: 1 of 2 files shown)


Levovar:
As usual and should have already anticipated after this many years, nobody actually gives a danm about CNI spec: Calico puts wrong interface name into its result, and doesn't do indexing on its own at all
And containerd treats "Interfaces" as mandatory, but is actually prepared to do indexing on its own (I guess they also realized somebody need to streamline the mess in code, otherwise nothing will ever work)

so, at the end of the day: eh. We just make sure the interface / IPs belonging the K8s Service Network are first in the CNI result, and let CRIs deal with the rest.
Seems to work, containerd recognizes the Calico V4/V6 addresses as PodIPs with this code when the Pod had 2 dual-stakc interfaces, one Calico and 1 IPVLAN

Levovar:
So for this CNI result:
`2021/05/20 16:07:56.421615 {
 "interfaces": [
  {
   "name": "cali5823a3ba4ac"
  },
  {
   "name": "eth1",
   "sandbox": "f128ffec7b20430d6e3cda7d576affc530217830b784501ea513a1e0093a1d3d"
  }
 ],
 "ips": [
  {
   "version": "4",
   "address": "192.168.180.222/32"
  },
  {
   "version": "6",
   "address": "fc00:caa5::14c1:6948:7f8f:d810/128"
  },
  {
   "version": "4",
   "address": "99.98.10.1/24"
  }
 ],
 "dns": {}
}`

these IPs were recognized by K8s:
`IP:                   192.168.180.222
IPs:
  IP:           192.168.180.222
  IP:           fc00:caa5::14c1:6948:7f8f:d810`

TothFerenc:
I could test it again by myself. It works fine. Ready to be merged. Thanks!

MikeZappa87:
@Levovar I am one of the CNI && containerd maintainers. I am curious about your thoughts on the spec and issues you have noticed.

Levovar:
> @Levovar I am one of the CNI && containerd maintainers. I am curious about your thoughts on the spec and issues you have noticed.

generally speaking or specifically related to this issue? :)

MikeZappa87:
> > @Levovar I am one of the CNI && containerd maintainers. I am curious about your thoughts on the spec and issues you have noticed.
> 
> generally speaking or specifically related to this issue? :)

I can discuss both :-) Please reach out .... Michael.Zappa@gmail.com or slack @Michael Zappa

PR #268, opened 2022-04-03
Updated the API domain to avoid problems with the last version of Kubernetes

### LINKED ISSUE #242: DANM API domain needs to be changed to comply with K8s API domain policy
K8s community implemented and released the following PR with K8s 1.19: https://github.com/kubernetes/enhancements/pull/1111

The TL;DR version of the Pr is that the API group used in the DANM CRDs unknowingly violates Kubernetes community policies, and therefore now must be changed so it doesn't "look like" it is an official K8s API.
Until this PR is implemented DANM is effectively not compatible with K8s versions 1.19+, because whenever the user would try and create a DANM API object the following error will be thrown:
```
Error from server (Invalid): error when creating "/var/lib/caas/crds/DanmEp.yaml": CustomResourceDefinition.apiextensions.k8s.io "danmeps.danm.k8s.io" is invalid: [spec.versions[0].schema.openAPIV3Schema: Requir
ed value: schemas are required, metadata.annotations[api-approved.kubernetes.io]: Required value: protected groups must have approval annotation "api-approved.kubernetes.io", see https://github.com/kubernetes/en
hancements/pull/1111] 
```

alejandrojnm:
**What type of PR is this?**
>
> bug
> cleanup
> design
> documentation
> failing-test

feature

Fixes #

fixes #242

I updated the API domain to avoid problems with the last version of Kubernetes, I already tested in a cluster and work as expected with the new domain `danm.io`

```release-note
NONE
```

### DIFF
--- a/crd/apis/danm/register.go
+++ b/crd/apis/danm/register.go
@@ -1,5 +1,5 @@
 package danm
 
 const (
-	GroupName = "danm.k8s.io"
+	GroupName = "danm.io"
 )
--- a/crd/apis/danm/v1/doc.go
+++ b/crd/apis/danm/v1/doc.go
@@ -2,5 +2,5 @@
 // +k8s:deepcopy-gen=package
 
 // Package v1 is the v1 version of the API.
-// +groupName=danm.k8s.io
+// +groupName=danm.io
 package v1
--- a/crd/client/clientset/versioned/typed/danm/v1/danm_client.go
+++ b/crd/client/clientset/versioned/typed/danm/v1/danm_client.go
@@ -33,7 +33,7 @@ type DanmV1Interface interface {
 	TenantNetworksGetter
 }
 
-// DanmV1Client is used to interact with features provided by the danm.k8s.io group.
+// DanmV1Client is used to interact with features provided by the danm.io group.
 type DanmV1Client struct {
 	restClient rest.Interface
 }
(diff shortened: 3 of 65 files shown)


Levovar:
this in itself is good, (looks like Travis is back to working order, yaay) but do note we have another repository hosting some utility functions which also somewhat depend on the correct API path: https://github.com/nokia/danm-utils
Particularly Cleaner is strongly proposed to be always deployed in production alongside the DANM meta, so in case you made this change because you use DANM in prod take note, you might want to extend teh scope to at least Cleaner as well

Levovar:
also note that this change in itself will not be enough to deploy DANM in newere K8s environments. DANM CRDs also need to be upgraded to GA V1 CRD API from v1beta1
the fix in this PR definitely solves one dependency of this upgrade, but if I recall correctly other API regulations might be also there

alejandrojnm:
Hi @Levovar, yes we know we need to update the existing DANM network to the new API, so we think we can use the danm-utils to do this, what did you think ??

Levovar:
not exactly sure what is the correlation between the two topics, in my mind they re separate
Issue1: the DANM API group name needs to be updated in danm-utils repo as well
Issue2: the CRD tAPI version in the DANM CRD templates need to be updated in both repos (was not part of this review) from v1beta to v1. Referring to these: https://github.com/nokia/danm/blob/master/integration/crds/production/ClusterNetwork.yaml#L1

BTW I just noticed you opened youe pull request against my branch, but not against master. I guess we still need a PR to rebase the changes to master

PR #269, opened 2022-05-03
Update the domain for danm

alejandrojnm:
**What type of PR is this?**
>
> bug
> cleanup
> design
> documentation
> failing-test
feature

Compatibility with the new versions of Kubernetes

Fixes #

I modify the client created by `client-gen` to add the Namespace to the Update function

```release-note
Now the new API is `danm.io/v1`, also the CRD is created using `apiextensions.k8s.io/v1` and not `apiextensions.k8s.io/v1beta1`
```


integration/crds/lightweight/DanmEp.yaml
```diff
@@ -1,15 +1,104 @@
...
+                    type: string
+                  proutes:
+                    additionalProperties:
+                      type: string
+                    type: object
+                  proutes6:
+                    additionalProperties:
+                      type: string
+                    type: object
+                required:
```
Levovar, line 61: not all interfaces have both V4 and V6 addresses etc
let's not unnecessairly add restrictions to an API which is only wirtten by code anyway, as it could result in failures in some edge cases

integration/crds/lightweight/DanmNet.yaml
```diff
@@ -1,116 +1,126 @@
...
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            properties:
+              AllowedTenants:
```
Levovar, line 35: AllowedTenants only exists on the ClusterNetwork API

integration/crds/lightweight/DanmNet.yaml
```diff
@@ -1,116 +1,126 @@
...
+                  allocation_pool:
+                    description: subset of the IPv4 subnet from which IPs can be allocated
+                    properties:
+                      end:
+                        type: string
+                      lastIp:
+                        type: string
+                      start:
+                        type: string
+                    required:
```
Levovar, line 60: dont add backward incompatble API changes not present on the original IP: allocation pools can be unbound in either direction

integration/crds/lightweight/DanmNet.yaml
```diff
@@ -1,116 +1,126 @@
...
+                    type: integer
+                  vlan:
+                    description: the VLAN id of the VLAN interface created on top
+                      of the host device
+                    type: integer
+                  vxlan:
+                    description: the vxlan id on the host device (creation of vxlan
+                      interface)
+                    type: integer
+                required:
```
Levovar, line 109: same

integration/crds/lightweight/DanmNet.yaml
```diff
@@ -1,116 +1,126 @@
...
+                  alloc:
+                    description: bit array of tracking address allocation
+                    type: string
+                  alloc6:
+                    description: bit array tracking IPv6 allocations
+                    type: string
+                  allocation_pool:
+                    description: subset of the IPv4 subnet from which IPs can be allocated
+                    properties:
+                      end:
```
Levovar, line 54: all the semantic checks like IP regex and integer range requirements were cut, any particular reason for this? the API semantics should not be changed, just updated

integration/crds/production/ClusterNetwork.yaml
```diff
@@ -1,120 +1,126 @@
...
 metadata:
-  name: clusternetworks.danm.k8s.io
+  name: clusternetworks.danm.io
 spec:
   scope: Cluster
-  group: danm.k8s.io
-  version: v1
+  group: danm.io
   names:
     kind: ClusterNetwork
```
Levovar, line 9: same comments as DanmNet, backward incompatible mandatory requirements added, validation rules missing

integration/crds/production/TenantNetwork.yaml
```diff
@@ -1,116 +1,125 @@
...
 metadata:
-  name: tenantnetworks.danm.k8s.io
+  name: tenantnetworks.danm.io
 spec:
   scope: Namespaced
-  group: danm.k8s.io
-  version: v1
+  group: danm.io
   names:
     kind: TenantNetwork
```
Levovar, line 9: same comments as DanmNet, backward incompatible mandatory requirements added, validation rules missing

pkg/admit/netadmit.go
```diff
@@ -1,19 +1,21 @@
...
+	"reflect"
+	"strings"
+	"time"
+
+	danmtypes "github.com/nokia/danm/crd/apis/danm/v1"
+	danmclientset "github.com/nokia/danm/crd/client/clientset/versioned"
+	"github.com/nokia/danm/pkg/confman"
+	"github.com/nokia/danm/pkg/metacni"
+	"k8s.io/api/admission/v1beta1"
+	// "k8s.io/kubernetes/cmd/kubeadm/test/kubeconfig"
```
Levovar, line 18: delete commented out code from final version

integration/crds/lightweight/DanmEp.yaml
alejandrojnm: I will keep only this tow
```yaml
required:
- MacAddress
- Name
```

alejandrojnm: unless you wanna remove `required` section

integration/crds/lightweight/DanmNet.yaml
alejandrojnm: you can explain this ???

integration/crds/lightweight/DanmNet.yaml
alejandrojnm: you can explain this too ???

integration/crds/lightweight/DanmNet.yaml
Levovar: you designated parameters as mandatory which weren't in the original API spec, therefore you made a backward incompatible change during the upgrade

integration/crds/lightweight/DanmNet.yaml
Levovar: you removed validation rules which were there in the original API spec, therefore you made a backward incompatible change during the upgrade

integration/crds/lightweight/DanmEp.yaml
Levovar: what I want is the API spec to remain semantics the **exact same** during the upgrade. I'm not really sure based on what you are deciding to remove some of the old rules, and add some previously non-existing designations?

Levovar:
1: is your commit message meant to imply you consider the comments resolved? if yes: they are not. I really fail to understand why do you keep tinkering with the existing attribute structure
you must keep everything as is. literally. go through the original YAML files attribute by attribute, and retain the exat same structure, with the exact same validation rules, without modifying them, or adding anything else to the spec
if for some reason this cannot be done explicitly state the reason attribute-by-attribute why modifications are required

2: so if I understand the context of client go related changes the only manual change you made here is the "Namespace" addition to the TenantConfig client's "Create" and "Update" operation
if I compare those with the code the same library generates for another namespaced API, for example for DanmNet we can see these Namespace additions are there by default: https://github.com/nokia/danm/blob/master/crd/client/clientset/versioned/typed/danm/v1/danmnet.go#L115
this makes me think the client generator code is actually **good**, and our instructions are bad! Take a look at the following instruction set for TenantConfigL:
https://github.com/nokia/danm/blob/master/crd/apis/danm/v1/types.go#L116
if you compare this to the designation of DanmNet you can see the order of the instructions is diferent. maybe that's an issue?
but more importantly if you compar the TenantConfigList instruction set:
https://github.com/nokia/danm/blob/master/crd/apis/danm/v1/types.go#L132
we told the client gen to generate the list nonNameSpaced. in all the other cases (except ClusterNetworkList) this designation is not there
TL;DR let's try to 1: switch the instruction order above TenantConfig struct to match the same order used in the case of other APIs 2: remove the "nonNameSpaced" designation from TenantConfigList so it matches the other APIs
then let's regenerated the code with these changes. I hope with these corrections the generated code will be okay and we don't need to manually fix it

Levovar:
@alejandrojnm looks good to me now! only thing left is to fix CI, and see if it works in a real environment

CI complains that your branch still has the manual modification in the TenantConfig client code, but with the new modifications you should be able to revert it, it should work with the baseline generated code. please test it out, and then we can merge

Levovar:
@alejandrojnm looks good to me! are you still finalizing or testing it, or can I go ahead and merge?

alejandrojnm:
I will test all tomorrow to see if work as expected, I will let you know, in that way you can merge

Levovar:
@alejandrojnm any updates?

alejandrojnm:
I will try to test this week, I have some tasks pending, sorry for the delay

alejandrojnm:
Hi @Levovar we test this in kubernetes 1.24 and work, I will send all the test env when we test all, but is working 🎉

PR #273, opened 2026-07-09
WE ARE SO BACK

Levovar:
Nokia first gave up on the project, then on the author but now we are both back, yeeee haaaaw. Or something like that

Let's start by changing the project to reference the new, correct, and most importantly independent organization it now belongs to!
Then if we are still feeling cute might also update the upstream dependencies later... IDK


Levovar:
Giga version bumps all over the place, we have now arrived to 2026

